PR #8, opened 2016-06-28
Extract WebIDL from respec and bikeshed docs

### LINKED ISSUE #3: Extract IDL elements defined in / referenced by a spec
Given the URL of a spec, extract:

1. the list of IDL interfaces and other IDL elements that the spec defines and that may be cross-referenced by other specifications.
2. the list of IDL elements that the spec references (e.g. `EventTarget` or other base classes)

This issue may perhaps need to be subdivided into issues that address specs written with a particular spec authoring tool.

dontcallmedom:
close #3

### DIFF
--- a/extract-webidl.js
+++ b/extract-webidl.js
@@ -1,0 +1,56 @@
+var jsdom = require('jsdom');
+
+function extract(url, cb) {
+    jsdom.env(url, [],
+              function(err, window) {
+                  if (err) return cb(err);
+                  var generator = window.document.querySelector("meta[name='generator']");
+                  if (generator && generator.content.match(/bikeshed/i)) {
+                      return extractBikeshedIdl(window.document, cb);
+                  }
+                  if (window.document.getElementById('respecDocument')) {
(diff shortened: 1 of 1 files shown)


extract-webidl.js
```diff
@@ -0,0 +1,56 @@
...
+    jsdom.env(url, [],
+              function(err, window) {
+                  if (err) return cb(err);
+                  var generator = window.document.querySelector("meta[name='generator']");
+                  if (generator && generator.content.match(/bikeshed/i)) {
+                      return extractBikeshedIdl(window.document, cb);
+                  }
+                  if (window.document.getElementById('respecDocument')) {
+                      return extractRespecIdl(window.document, cb);
+                  }
```
tidoust: So, that only works for generated version of ReSpec documents.

See code in https://github.com/tidoust/reffy/commit/4873d9d2c3289b3153db358ece7bd76051e1d775 for a way to make JSDom run ReSpec and attach this code to after ReSpec is done generating the document.

Feel free to merge as-is, I'll make the update afterwards.

PR #9, opened 2016-06-28
Extract information on names and dependencies from WebIDL

dontcallmedom:
Builds on #8

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "dependencies": {
-    "jsdom": "^9.3.0"
+    "jsdom": "^9.3.0",
+    "webidl2": "^2.0.11"
   }
 }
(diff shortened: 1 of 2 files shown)


dontcallmedom:
(the pull request was made on the branch webidl-parser but obviously we want to merge it to master if ok with it)

tidoust:
That looks good to me. I note that this somewhat fails for the [Presentation API](https://www.w3.org/TR/presentation-api/), because the spec contains two `partial interface Presentation` definitions, and only the second one appears in the resulting tree. Not much we can do about it, each partial interface is to be implemented by a different conformance class, which cannot be guessed automatically. The result remains correct because the first definition does not introduce additional external dependencies.

Anyway, I'll merge this on master.

tidoust:
Merge done manually.

Issue #6, opened 2016-06-27
Extract normative/informative references from TR specs 


tidoust:
Given the URL of a document published in TR space, extract the list of normative/informative references that this document lists and produce a JSON structure with the list of references.

The main difficulty is that no single recipe will work across specifications, so the tool will likely have to maintain a set of per-spec rules to follow to create a useful list.

PR #10, opened 2016-06-28
Extract more references (from other generators)

dontcallmedom:
This makes it possible to extract all references from the document listed at https://www.w3.org/standards/techs/js#w3c_all with the exception of serviceworkers (but works with editors draft) and selection-api (but that's because it wrongly classifies its references as purely informative).

### DIFF
--- a/parse-references.js
+++ b/parse-references.js
@@ -74,14 +74,19 @@ function nextTag(node, name) {
     return nextSibling;
 }
 
-function extractReferencesFromList(referenceList) {
-    return [].map.call(referenceList.querySelectorAll("dt"), function(dt) {
+function extractReferencesFromList(referenceList, type) {
+    var defaultRef = [], informativeRef = [];
+    [].forEach.call(referenceList.querySelectorAll("dt"), function(dt) {
         var ref = {};
         ref.name = dt.textContent.replace(/[\[\] \n]/g, '');
         var desc = nextTag(dt, "dd");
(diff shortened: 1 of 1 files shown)


dontcallmedom:
I guess this addresses #6

parse-references.js
```diff
@@ -22,15 +22,26 @@ var extractionRules = {
...
 
 function extract(url, cb) {
     jsdom.env({
         url: url,
         features: {
             FetchExternalResources: ['script'],
-            ProcessExternalResources: ['script']
+            ProcessExternalResources: ['script'],
+            // only load respec, to avoid jsdom bugs
+            SkipExternalResources: /^((?!respec).)*$/
```
tidoust: Should be applied to extract-webidl.js as well, I suppose.

dontcallmedom: indeed; in general, we probably need a core spec-util module to DRY these modules.

tidoust:
I made some tweaks to improve readability a bit (from some definition readability, that is ;)). I'll merge now and create a "util" module.

tidoust:
Are there specs that both contain an informative list and flag individual references as "non-normative"?

dontcallmedom:
how would I know? that's what defensive programming is for >:)

tidoust:
OK, well, your defensive code would have failed too ;) (you forced that type to "normative" for informative references extraction as well). I'll add something back.

dontcallmedom:
nah, I think you can simply ignore it, I'm just being annoying :)

tidoust:
Too late :)

PR #12, opened 2016-06-29
Gather dependencies at the interface level

### LINKED ISSUE #11: List dependencies at the interface level
PR  #9 does the job to list external dependencies of a particular specification, which is what we're aiming at.

Could it be useful to also list dependencies at the "interface" level, typically the first level that appears under `idlNames`? It would allow to create the interface visualization that could prove more useful (especially if it's in 3D and viewed with a VR headset ;)).

At that level, no need to make a distinction between internal and external dependencies. On top of my head, I would only include direct dependencies but perhaps it would be better to pull nested dependencies as well.

dontcallmedom:
close #11

### DIFF
--- a/parse-webidl.js
+++ b/parse-webidl.js
@@ -3,7 +3,7 @@ var WebIDL2 = require("webidl2");
 function parse(idl, cb) {
     var idlTree;
     var jsNames = {constructors: {}, functions: {}, objects:{}};
-    var idlNames = {};
+    var idlNames = {_dependencies: {}};
     var idlExtendedNames = {};
     var localNames= {};
     var externalDependencies = [];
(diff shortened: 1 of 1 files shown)


tidoust:
That looks good. I guess I would have merged the dependencies with the actual IDL names as in the following extract, instead of in a separate "_dependencies" bucket, but the most important point right now is to have the info somewhere easily accessible.

```json
{
  "idlName": {
    "Presentation": {
      "type": "interface",
      ...
      "dependencies": [
        "PresentationRequest",
        "PresentationReceiver"
      ]
    }
  }
}

dontcallmedom:
for the record, I initially thought I would do as you describe. The reason I went with a different hierarchy is that it would have been otherwise messy for partial interfaces: I would have had to create an entry in idlNames for them, but that entry itself would not match what other entries in the tree had (i.e. the WebIDL AST).

PR #13, opened 2016-06-29
Add spec crawler

dontcallmedom:
Given a list of URLs, it fetches the references and IDL from the said URLs in a gigantic JSON tree.


tidoust:
That looks good. One small improvement that we might want to do "when time allows" is to avoid loading the spec twice to parse references and WebIDL.

That said, we may have to load two different URLs for specs that are too large.

PR #14, opened 2016-06-29
 	Crawl editors draft when available

dontcallmedom:
This uses the w3c API, so needs a key configured in a config.json file with key 'w3cApiKey'


tidoust:
That looks good but aren't we losing the link between the initial URL (and/or the shortname) and the URL of the editor's draft?

We'll need that mapping for the normalization exercise. Happy to do it.

dontcallmedom:
I've reworked loadSpecification (and its usage) to avoid multiple loading/parsing of the same spec, and made the crawler work on chunk of the list - it makes it more resilient, but still core-dumps when the HTML spec is in the list :(

dontcallmedom:
(the core-dump is due to memory exhaustion, so I suspect there must be a memory leak somewhere, but I'm not sure how to debug it)

tidoust:
That all looks good. I see how things could be improved, but that's all time permitting, and time does not easily permit in practice.

PR #31, opened 2017-06-26
Use respec to generate DOM from respec documents

### LINKED ISSUE #28: ReSpec timeout
When running reffy, respec times out when fetching references (afaict)

dontcallmedom:
Also upgrade to new jsdom API (>10)
close #28

### DIFF
--- a/util.js
+++ b/util.js
@@ -230,7 +230,7 @@ function loadSpecification(url) {
             // ReSpec doc
             if (doc.head.querySelector("script[src*='respec']")) {
                 // this does another network fetch :(
-                return respecWriter(url, '/dev/null', {}, 20000).then(function(html) {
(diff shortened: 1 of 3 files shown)


util.js
```diff
@@ -224,79 +225,38 @@ function fetch(url, options) {
...
-                                    .then(window => resolve(window))
-                                    .catch(err => reject(err));
-                            }
-                            return;
+            const {window} = new JSDOM(html, {url: response.url});
+            const doc = window.document;
+            // ReSpec doc
+            if (doc.head.querySelector("script[src*='respec']")) {
+                // this does another network fetch :(
+                return respecWriter(url, '/dev/null', {}, 20000).then(function(html) {
```
tidoust: `/dev/null` does not mean anything on Windows and makes the code crash. Using an empty string `''` seems to do what you want.

Issue #29, opened 2017-06-26
Canonicalize URLs


dontcallmedom:
When detecting references and IDL definitions, we need to use canonical URLs for a given spec, taking into account:
- multiple "levels" of a spec
- multiple versions (WD, CR, REC, PER, 2nd ed Rec)
- TR vs ED
- WHATWG vs W3C (where sometimes the scope of a given spec are not a 1:1)

dontcallmedom:
presumably, `spec-equivalents.json` can help some here now.

Issue #27, opened 2017-06-26
Build spec list (semi-?)automatically


dontcallmedom:
Starting from a manually curated list of specs is not ideal, since new specs emerge all the time, and old specs get replaced by newer versions.

One approach is to derive the list automatically from https://platform.html5.org/ and/or https://www.w3.org/standards/techs/js#w3c_all ; another might be to extract specs from caniuse or chrome platform status.

We would probably want a blacklist of specs to ignore

tidoust:
Doing some house-keeping here. We have a semi-automatic mechanism in place to detect new specs. That could be improved, for sure, but seems good enough for now. Closing this issue as a result.

PR #32, opened 2017-06-26
Find relevant specs unknown to the system

dontcallmedom:
part of #27
Needs lots of canonicalization (see #29) to reduce false positives


dontcallmedom:
@tidoust I think this is now ready - I have added all the specs that my `find-spec.js` script reported (and marked the others as to be ignored)

A further improvement would be to run that script on a regular basis and automatically raise an issue when its result is not an emtpy array

tidoust:
@dontcallmedom That looks mostly good to me, apart from the fact that "results.json" was not a parameter. See updated PR.

Also, I found your Promise logic a bit convoluted, so I tried to "simplify" that a bit. Result should be the same. Can you have a look?

PR #33, opened 2017-06-26
Add a mechanism to normalize WebIDL1 into WebIDL2

dontcallmedom:


tidoust:
Update already incorporated in #32, closing without merging.

PR #42, opened 2017-06-27
Split spec list to create different views based on overlapping W3C/WH…

dontcallmedom:
…ATWG specs

This should help find inconsistencies in a specific set of specs


tidoust:
@dontcallmedom I added a simple mechanism to the crawler to reference a list of specs from within another list of specs, and moved all common specs to `specs-common.json` as a result, with URLs sorted by alphabetic order (well origin type then alphabetic order).

That common list of specs is now referenced by both `specs-w3c.json` and `specs-whatwg.json`.

PR #45, opened 2017-06-28
Handle parsing types of multi-type generics

dontcallmedom:
e.g. record<DOMString,Foo> - so far, the code assumed generics only took one parameter

### DIFF
--- a/parse-webidl.js
+++ b/parse-webidl.js
@@ -270,7 +270,7 @@ function parseType(idltype, idlNames, externalDependencies, contextName) {
     if (isString(idltype)) {
         idltype = { idlType: 'DOMString' };
     }
-    if (idltype.union) {
+    if (idltype.union || (idltype.generic && Array.isArray(idltype.idlType))) {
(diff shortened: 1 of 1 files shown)


parse-webidl.js
```diff
@@ -282,6 +282,7 @@ function parseType(idltype, idlNames, externalDependencies, contextName) {
                           "RegExp", "Error", "DOMException", "ArrayBuffer", "DataView", "Int8Array", "Int16Array", "Int32Array", "Uint8Array", "Uint16Array", "Uint32Array", "Uint8ClampedArray", "Float32Array", "Float64Array",
                           "ArrayBufferView", "BufferSource", "DOMTimeStamp", "Function", "VoidFunction"];
     if (wellKnownTypes.indexOf(idltype.idlType) === -1) {
+            console.error(idltype);
```
tidoust: That seems like a leftover from debug. If not, the message should be more explicit.

dontcallmedom: fixed

Issue #26, opened 2017-02-13
List specs that depend on a given spec


chaals:
A common problem when messing with some API is to know who is a consumer.

In general, that's pretty much unanswerable. But it would be helpful if reffy would at least provide a simple list of specs that depend on X, so if you are changing X you can let the spec people who might be upset know…

tidoust:
@chaals First, thanks for the suggestion! @dontcallmedom and I managed to get some work done on Reffy this week and Reffy can now produce a dependencies report. Closing this as a result.

By definition, Reffy will only report on incoming links to a specification from specifications it knows about. In particular, note that we have retained an IDL focus for now: the crawl does not include non-IDL specs (many CSS specs are missing for instance).

PR #48, opened 2017-06-28
First stab at dependencies report

tidoust:
This PR addresses #26 and adds a dependencies report to the study tool, triggered when the console parameter is "dep". That dependencies report lists specs that got crawled. For each of them, the report lists the specs that reference it either normatively or informatively.

The report could be further improved. It could for instance clarify what version of the spec the reference targets (ED, version number when available, dated spec, generic TR spec, etc.). It could also list the references that are currently missing but that should exist (one of the anomalies that we report in other reports).

### DIFF
--- a/study-specs.js
+++ b/study-specs.js
@@ -22,18 +22,19 @@ function processReport(results) {
     // TODO: we may end up with different variants of the WebIDL spec
     var WebIDLSpec = results.find(spec => (spec.shortname === 'WebIDL-1')) || {};
 
-    return results
-        .sort((a, b) => {
-            var A = a.title.toUpperCase();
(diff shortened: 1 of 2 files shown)


study-specs.js
```diff
@@ -22,18 +22,19 @@ function processReport(results) {
...
+        var B = b.title.toUpperCase();
+        if (A < B) {
+            return -1;
+        }
+        if (A > B) {
+            return 1;
+        }
+        return 0;
+    });
+
```
dontcallmedom: or more simply
```js
var sortedResults = results.sort((a,b) => a.title.toUpperCase().localeCompare(b.title.toUpperCase()));
```

PR #49, opened 2017-06-28
Report ref URLs that are inconsistent with URLs used in links

tidoust:
Thanks to equivalence and crawling, we can now mostly easily map between different versions of the same spec (ED, dated spec, latest published version, previous version, etc.). However, while this mapping helps spotting missing references to specs, doing it automatically can also hide possible inconsistencies in a spec when the body of the spec may link to a particular version of a spec whereas the final reference is to another version.

Reports now include a section that alerts about such inconsistencies.

Internally, canonicalization now has two levels:
1. by default, it returns the "same" document
2. if requested, it may handle dated specs and take equivalences into account


tidoust:
Merging this for now to continue work. We can always revisit some of it later on.

PR #61, opened 2017-06-30
Add a mechanism to filter specs per anomaly to the HTML report

tidoust:


html/reffy.js
```diff
@@ -1,13 +1,18 @@
 window.onload = function () {
+  const $ = (el, selector) =>
+    Array.prototype.slice.call(el.querySelectorAll(selector), 0);
+
```
dontcallmedom: nice :)

Issue #66, opened 2017-09-06
ElementTraversal is now part of DOM


dontcallmedom:
cf https://github.com/heycam/webidl/issues/365#issuecomment-327420054

tidoust:
PR prepared. Is there an existing plan to obsolete the ElementTraversal spec on our end?

dontcallmedom:
https://github.com/w3c/WebPlatformWG/issues/96

PR #70, opened 2017-09-08
Drop the ElementTraversal specification from crawl results

tidoust:
The ElementTraversal Recommendation is now incorporated in the DOM spec (both the W3C and WHATWG version) and the ElementTraversal spec is not referenced from any other spec.

Executes order #66.

### DIFF
--- a/specs-common.json
+++ b/specs-common.json
@@ -1,7 +1,6 @@
 [
     "https://www.w3.org/TR/CSP3/",
     "https://www.w3.org/TR/DOM-Parsing/",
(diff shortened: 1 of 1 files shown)


dontcallmedom:
I was wondering if we needed to add stuff to the `-equivalent` files - but I gather that we don't because no other spec references it, right?

tidoust:
Right, no other spec references ElementTraversal so it won't appear anywhere if we drop it from the list of specs being crawled, and as such, there is no need to make the mapping to DOM explicit.

PR #82, opened 2018-01-20
Retry network fetches a few times to work around transient errors

tidoust:
For some reason, fetching the SVG 2 spec fails with a "socket hang up" from time to time. The fetch function now tries again after a few seconds when such errors happen. Hopefully, that will make Reffy a bit more resilient to transient network issues.

Not related, but the result of the check on pending fetches was actually not used, meaning Reffy could fetch the same URL multiple times at once. Fixed in this commit as well.

### DIFF
--- a/util.js
+++ b/util.js
@@ -96,6 +96,10 @@ function fetch(url, options) {
         }
     }
 
+    function addPendingFetch(url, promise) {
+        pendingFetches[url] = promise;
+    }
+
     function readHeadersFromCache() {
         return new Promise((resolve, reject) => {
             fs.readFile(cacheHeadersFilename, 'utf8', function (err, data) {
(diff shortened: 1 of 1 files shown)


util.js
```diff
@@ -192,23 +196,44 @@ function fetch(url, options) {
...
+            return (remainingAttempts <= 0) ?
+                baseFetch(url, options) :
+                baseFetch(url, options).catch(err => {
+                    if (!options.nolog) console.warn('Fetch attempt failed: ' + url);
+                    return new Promise((resolve, reject) => {
+                        setTimeout(function () {
+                            fetchWithRetry(url, options, remainingAttempts - 1)
+                                .then(resolve)
+                                .catch(reject);
+                        }, 2000 + Math.floor(Math.random() * 8000));
```
dontcallmedom: out of curiosity, why the random length for the timeout?

tidoust: I guess I cannot make up my mind as to how long the code should wait before retrying, because I don't know the cause of the "socket hang up" message. I thought it would make sense to introduce an arbitrary delay not to hit the server again at specific intervals. But now that you make me think about it, I don't quite see what that could bring.

Typically, I do not know where the "svgwg.org" Web site is hosted, but perhaps it's running on a virtual machine that goes to sleep after some time in the absence of requests and that first needs to restart when a new request comes in, which could make it timeout on the first request. A random length won't help there, we just need to retry after 5 seconds. I'll drop the random part.

tidoust: Ah, I see you already merged this with the random counter. Well, no big deal.

PR #93, opened 2018-03-19
Add support for exporting crawled IDL fragments

dontcallmedom:
For https://github.com/tidoust/reffy-reports/issues/3


crawl-specs.js
```diff
@@ -325,8 +343,19 @@ function crawlList(speclist, crawlOptions) {
...
+      spec =>
+        new Promise((resolve, reject) => {
+          if (spec.idl.idl) {
+            fs.writeFile(pathlib.dirname(path) + '/idl/' + getShortname(spec) + '.idl',
+                         spec.idl.idl,
+                         err => { if (err) return console.log(err); return resolve();});
+            delete spec.idl.idl;
+          } else resolve();
+        })))
+    .then(new Promise((resolve, reject) => {
```
tidoust: This does not seem to bug Node.js, but that is wrong. First parameter to `then` should be a `Function`, not a `Promise`, so this should rather be:
`.then(_ => new Promise((resolve, reject) => {`

(In practice, the code of that promise currently runs _before_ the previous `Promise.all` call and any data it could return when it resolves gets ignored, which has no consequence here, but not exactly what you'd expect from a chain of `then`....)
That still seems w

crawl-specs.js
```diff
@@ -325,8 +343,19 @@ function crawlList(speclist, crawlOptions) {
...
  */
 function saveResults(crawlInfo, crawlOptions, data, path) {
-    return new Promise((resolve, reject) => {
-        fs.readFile(path, function(err, content) {
+  return Promise.all(
+    data.map(
+      spec =>
+        new Promise((resolve, reject) => {
+          if (spec.idl.idl) {
+            fs.writeFile(pathlib.dirname(path) + '/idl/' + getShortname(spec) + '.idl',
```
tidoust: Would be good to create the `idl` folder when it does not exist.

dontcallmedom:
instead of doing the `mkdir` in the code, I did as for other directories and created them in the repo.

tidoust:
> instead of doing the mkdir in the code, I did as for other directories and created them in the repo.

Hmm. So, first, you did not ;) The `.gitkeep` files that I see in the PR are for the parent report folders (which were already preserved through a README file), not for the `idl` folders.

Second, this works for our own usage, but the command-line usage of `crawl-specs.js` becomes clumsy: it expects the filename of the report to create, but then the user would have to create an `idl` folder (which comes out of the blue) on her own, that seems bad.

I'm going to push an update to the PR that turns the second parameter of `crawl-specs.js` into a folder (to make it clearer that the tool will output a bunch of files in that folder), and that creates the `idl` folder in that folder if it does not exist.

tidoust:
@dontcallmedom Can you have a look at the update?

Also, do you want to address the problem of missing IDL files for some of the specs (e.g. WebAssembly, CSS Layout API) here or in a separate issue?

dontcallmedom:
Your changes make plenty of sense, indeed. I've added a shortname matcher for khronos specs, and I think we should merge as is; we can explore the missing specs separately.

Issue #90, opened 2018-03-16
Publish as NPM module


dontcallmedom:
I've re-used some of the existing libraries in a number of small projects (in particular, the ability to extract and parse webidl from specs) - this would be made easier if reffy (or a subset of it?) was made available as an npm module

tidoust:
Closing this issue as Reffy is now available as an [npm package](http://npmjs.com/package/reffy).

v1.1.1 is the first version that runs correctly as an NPM package (a minor revision to add an `reffy.js` CLI to pilot Reffy as an installed module, followed by a patch revision to fix remaining path resolution error in spec list files). I wouldn't be surprised if something is still not running as intended when Reffy is used as an NPM module. Potential problems to be raised as separate issues...

PR #96, opened 2018-03-29
Update package.json for publication as an NPM package

tidoust:
As requested in #90.

Add a package name (reffy), links to repo, names, license and list command-line files. Also updated JS files to add required shebang.

First version is 0.4.0 because because.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "reffy",
-  "version": "0.4.0",
+  "version": "1.0.0",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -1,4 +1,36 @@
 {
+  "name": "reffy",
+  "version": "0.4.0",
```
dontcallmedom: I'd go with 1.0 - it's been in use for a while now

also, I think we should commit to use semver - not sure if that's something that is exposed in package.json

tidoust: Following the semver format is required in package.json, but I don't think there's any way to assert in that file that we commit to follow semver conventions. I'll adjust the version number, merge and publish.

PR #99, opened 2018-05-07
Remove https://www.w3.org/TR/selectors-api/

foolip:
See https://github.com/tidoust/reffy-reports/issues/3#issuecomment-386557144


tidoust:
Thanks, @foolip.

FWIW, note I will actually keep the Selectors API spec in the W3C crawl and report: one of the goals of Reffy is to highlight inconsistencies that we have, typically here the fact that the spec needs to be obsoleted.

PR #106, opened 2018-05-15
Have ReSpec run in JSDOM, stop using RespecDocWriter

tidoust:
Reffy needs to run ReSpec on Editor's Drafts. It relied on RespecDocWriter until now, which uses Puppeeter, which in turn relies on a headless version of Chrome. That requires downloading 100s of MB of dependencies and is a tad heavy for the task at hands, especially as we consider running Reffy as an HTTP service.

With this update, RespecDocWriter is no longer being used. Instead, Reffy makes ReSpec run in JSDOM, which is much more lightweight (and JS-based).

Problem is that JSDOM does not yet support enough features to run ReSpec. Concrete issues are documented in:
https://github.com/w3c/respec/issues/1469#issuecomment-388560878

This update monkey-patches JSDOM's code to add good-enough implementations of missing features, as well as to regain control over network requests so that all network requests go through our cache. It also monkey-patches ReSpec to:

1. drop a couple of modules that cannot run because they use non-implemented features of JSDOM that cannot be easily patched
2. update the code of a couple of ReSpec dependencies for the same reason

The whole thing remains fragile: updates to ReSpec or to its dependencies may well break the regular expressions used to monkey patch the code. Most of the monkey-patching should resist minor changes. A couple of them may not though. I suggest we try this for some time and see whether that proves too hard to maintain.

### DIFF
--- a/jsdom-monkeypatch.js
+++ b/jsdom-monkeypatch.js
@@ -1,0 +1,308 @@
+/**
+ * Reffy runs ReSpec in JSDOM to avoid having to download an run a complete
+ * headless browser.
+ *
+ * Problem is JSDOM does not yet support a number of features that ReSpec needs.
+ * Also, JSDOM does not allow interception of network requests anymore.
+ *
+ * This code monkey-patches JSDOM (and ReSpec) code before other modules make
+ * use of them, so that ReSpec can run in JSDOM and so that network requests
+ * use our fetch implementation.
+ *
+ * This module needs to be required before any other module that make use of
+ * JSDOM.
+ *
+ * The whole thing is PRETTY UGLY.
+ *
+ * NB: This code WILL LIKELY BREAK when switching to new versions of JSDOM or
+ * ReSpec.
+ */
+
+const resourceLoader = require('jsdom/lib/jsdom/browser/resource-loader');
+const Attr = require('jsdom/lib/jsdom/living/generated/Attr').interface;
+const Element = require('jsdom/lib/jsdom/living/generated/Element').interface;
+const fetch = require('./fetch.js');
+const { Headers, Request, Response } = require('node-fetch');
+
+resourceLoader.download = function (url, options, callback) {
+  // Restrict resource loading to ReSpec and script resources that sit next
+  // to the spec under test, excluding scripts of WebIDL as well as the
+  // WHATWG annotate_spec script that JSDOM does not seem to like.
+  // Explicitly whitelist the "autolink" script of the shadow DOM spec which
+  // is needed to initialize respecConfig
+  const respecUrl = 'https://www.w3.org/Tools/respec/respec-w3c-common';
+  function getUrlToFetch() {
+    let referrer = options.referrer;
+    if (!referrer.endsWith('/')) {
+      referrer = referrer.substring(0, referrer.lastIndexOf('/') + 1);
+    }
+    if (/\/respec[\/\-]/i.test(url.path)) {
+      //console.log(`fetch ReSpec (force latest version)`);
+      return respecUrl;
+    }
+    else if (/\.[^\/\.]+$/.test(url.path) &&
+        !url.path.endsWith('.js') &&
+        !url.path.endsWith('.json')) {
+      //console.log(`fetch not needed for ${url.href} (not a JS/JSON file)`);
+      return null;
+    }
+    else if ((url.pathname === '/webcomponents/assets/scripts/autolink.js') ||
+        (url.href.startsWith(referrer) &&
+          !(/annotate_spec/i.test(url.pathname)) &&
+          !(/expanders/i.test(url.pathname)) &&
+          !(/bug-assist/i.test(url.pathname)) &&
+          !(/dfn/i.test(url.pathname)) &&
+          !(/section-links/i.test(url.pathname)) &&
+          !(/^\/webidl\//i.test(url.pathname)))) {
+      //console.log(`fetch useful script at ${url.href}`);
+      return url.href;
+    }
+    //console.log(`fetch not needed for ${url.href}`);
+    return null;
+  }
+
+  let urlToFetch = getUrlToFetch();
+  if (!urlToFetch) {
+    return callback(null, '');
+  }
+  fetch(urlToFetch, options)
+    .then(response => response.text())
+    .then(data => {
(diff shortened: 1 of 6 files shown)


jsdom-monkeypatch.js
```diff
@@ -0,0 +1,308 @@
...
+ * use of them, so that ReSpec can run in JSDOM and so that network requests
+ * use our fetch implementation.
+ *
+ * This module needs to be required before any other module that make use of
+ * JSDOM.
+ *
+ * The whole thing is PRETTY UGLY.
+ *
+ * NB: This code WILL LIKELY BREAK when switching to new versions of JSDOM or
+ * ReSpec.
```
dontcallmedom: is there a simple test we can build to detect that early?

jsdom-monkeypatch.js
```diff
@@ -0,0 +1,308 @@
...
+      ["core/highlight", "core/list-sorter"]
+        .forEach(module => data = data.replace(
+          new RegExp('(define\\(\\s*"profile-w3c-common"\\s*,\\s*\\[[^\\]]+),\\s*"' + module + '"'),
+          '$1'));
+
+      // JSDOM's CSS parser does not quite like uncommon "@" rules
+      // so let's pretend they are just @media rules
+      // https://github.com/jsdom/jsdom/issues/2026
+      // (NB: this replacement is just for convenience, to avoid JSDOM reporting
+      // lengthy errors (including a full dump of the CSS) to stderr
```
dontcallmedom: I think in practice we're not interested in CSS parsing - cf https://github.com/jsdom/jsdom/issues/2005

jsdom-monkeypatch.js
```diff
@@ -0,0 +1,308 @@
...
+      data = data.replace(/@keyframes \S+? {/, '@media all {');
+      data = data.replace(/@supports \(.+?\) {/, '@media all {');
+
+      // Respec drops blank lines in Markdown, but marked.js actually
+      // needs them around <pre> tags, otherwise it produces really weird
+      // HTML (with <pre> and <p> intertwined). For some reason, this does
+      // not bother regular browsers. It does bother JSDOM though.
+      data = data.replace(/r\.createTextNode\("\\n"\)/, 'r.createTextNode("\\n\\n")');
+
```
dontcallmedom: maybe drop `r.` to make less fragile?

reffy.js
```diff
@@ -153,8 +153,12 @@ program
       }
     });
 
-    promise.then(_ => console.log('-- THE END -- '))
-      .catch(err => console.error(err));
+    return promise.then(_ => console.log('-- THE END -- '))
+      .catch(err => {
+        console.error('-- ERROR CAUCHT --');
```
dontcallmedom: s/UCH/UGH/

util.js
```diff
@@ -78,19 +45,47 @@ function fetch(url, options) {
...
         // Drop Byte-Order-Mark character if needed, it bugs JSDOM
         if (html.charCodeAt(0) === 0xFEFF) {
             html = html.substring(1);
         }
-        const {window} = new JSDOM(html, {url: responseUrl});
-        const doc = window.document;
+        const {window} = new JSDOM(html, {
+            url: responseUrl,
+            resources: 'usable',
+            runScripts: 'dangerously',
```
dontcallmedom: thrilling!

jsdom-monkeypatch.js
tidoust: Actually, there's another call to `createTextNode("\n")` somewhere else in the code which would be modified as well. I'll stick with that code for now, but will investigate upstreaming that change to ReSpec (after all, the HTML produced is wrong in regular browsers as well, it just happens not to have any bad consequence there)

util.js
tidoust: Yes, any interaction with dom is potentially dangerous, I've always said that :)

jsdom-monkeypatch.js
tidoust: For JSDOM, we control the switch to a new version, so running a crawl locally before pushing the version bump could be good-enough (but it would certainly be better to write a few test cases!).

For ReSpec, without test cases, it's a bit more difficult, because the latest version of ReSpec is used "on the fly". We could force the use of a specific version of ReSpec to make sure updates are explicit and remain under our control.

jsdom-monkeypatch.js
```diff
@@ -0,0 +1,308 @@
...
+      // (NB: this replacement is just for convenience, to avoid JSDOM reporting
+      // lengthy errors (including a full dump of the CSS) to stderr
+      data = data.replace(/@keyframes \S+? {/, '@media all {');
+      data = data.replace(/@supports \(.+?\) {/, '@media all {');
+
+      // Respec drops blank lines in Markdown, but marked.js actually
+      // needs them around <pre> tags, otherwise it produces really weird
+      // HTML (with <pre> and <p> intertwined). For some reason, this does
+      // not bother regular browsers. It does bother JSDOM though.
+      data = data.replace(/r\.createTextNode\("\\n"\)/, 'r.createTextNode("\\n\\n")');
```
Zirro: @tidoust We would be interested in fixing this in jsdom (or potentially in [parse5](https://github.com/inikulin/parse5)). Do you have a sample of the HTML which behaves differently between jsdom and other browsers?

Your list in https://github.com/w3c/respec/issues/1469#issuecomment-388560878 is very nice as well. It'd be great if you could file new issues if you find something not covered by the existing ones.

tidoust: Thanks @Zirro! I don't think there is any bug in jsdom, actually. What I get from jsdom is pretty much what I expect to get (wrong markup in, wrong markup out). The output I do not understand is the one in regular browsers... but I suspect that the wrong markup gets fixed by chance by ReSpec afterwards when it runs the `core/highlight` module (module which cannot run in jsdom). I ran out of time on this issue, so did not investigate further, but the real bug is in marked.js who produces wrong markup, not in jsdom.

I didn't file any issue in jsdom from the list because all of the current missing features either appear as real issues already, or are explicit pointed out as missing in code comments or commit messages. So that all seemed well-known already. Reffy now runs ReSpec in JSDOM (with the monkey-patching above), so we'll keep an eye on potential new problems that we may find!

PR #118, opened 2018-08-14
Include the spec title+URL in CSS JSON output

foolip:
This is to make it easier to understand the origin of the file
contents, and possibly for tools to generate boilerplate for test.


foolip:
Please be warned that this is entirely untested, I hope it works :)

foolip:
Thanks @tidoust, now waiting for the next update of https://github.com/tidoust/reffy-reports/commits/master :)

Would it be possible to trigger a reffy-reports update if the master branch of this repo changes? That'd be nice, although not particularly important.

PR #121, opened 2018-08-14
Add MediaStreamTrack Content Hints to list of specs

foolip:


foolip:
I see https://github.com/tidoust/reffy-reports/blob/master/whatwg/idl/mst-content-hint.idl, I guess that manually added from the result of trying this PR?

tidoust:
Yes, I just ran the crawl on my computer to create the right JSON file and put it into the report (putting an empty file would work too, it's just a way to tell the job that runs daily that updates to this file should be tracked from now on).

PR #132, opened 2018-11-16
Scrape Service Workers Nightly for WHATWG data set

foolip:
This is because v1 is a subset, e.g., FetchEvent:
https://w3c.github.io/ServiceWorker/v1/#fetchevent-interface
https://w3c.github.io/ServiceWorker/#fetchevent-interface

Notices because resultingClientId is missing in wpt/interfaces/.

### DIFF
--- a/src/specs/spec-equivalents.json
+++ b/src/specs/spec-equivalents.json
@@ -71,10 +71,6 @@
     "https://www.w3.org/TR/html-media-capture/": [
         "https://dev.w3.org/2009/dap/camera/"
     ],
-    "https://www.w3.org/TR/service-workers-1/": [
-        "https://w3c.github.io/ServiceWorker/",
-        "https://www.w3.org/TR/service-workers/"
-    ],
     "https://www.w3.org/TR/webvtt1/": [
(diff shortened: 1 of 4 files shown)


src/specs/spec-equivalents.json
```diff
@@ -71,10 +71,6 @@
     "https://www.w3.org/TR/html-media-capture/": [
         "https://dev.w3.org/2009/dap/camera/"
     ],
-    "https://www.w3.org/TR/service-workers-1/": [
```
tidoust: The equivalences listed here merely affect the way Reffy will look at references to answer the question: "This spec uses IDL terms defined in Service Workers, does it reference it normatively?" The exact URL used for the reference does not matter so much as the presence of the reference (although we could certainly add a test about URL consistency later on).

In other words, I would still keep these equivalences as is.

(I haven't checked whether these specific equivalences for Service Workers are still useful though).

foolip:
I've revert that change. @tidoust PTAL.

tidoust:
Thanks, rest is fine!

PR #137, opened 2019-01-17
Exclude IDL definitions that have an `exclude` class name

tidoust:
When Reffy extracts individual IDL definition snippets, it now skips over blobs defined with an `exclude` class name, and snippets that are defined in informative sections. The rules are identical to those defined in ReSpec.

Needs to be tested before merging (cannot test for now because I'm on a restricted network).

### DIFF
--- a/src/cli/extract-webidl.js
+++ b/src/cli/extract-webidl.js
@@ -93,6 +93,13 @@ function extractBikeshedIdl(doc) {
  * @return {Promise} The promise to get a dump of the IDL definitions
  */
 function extractRespecIdl(doc) {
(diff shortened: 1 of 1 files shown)


foolip:
Probably because of these changes, `dictionary RTCStats` was dropped in https://github.com/tidoust/reffy-reports/commit/85b22abe3b0f4c02973a7b37ff6980c82bb182fb. However, it was dropped for both specs...

foolip:
No, sorry, I misread the diff, both w3c/idl/webrtc-stats.idl and whatwg/idl/webrtc-stats.idl were modified, as they should be.

PR #139, opened 2019-01-29
Add missing valuespaces id and outline-line-style

pyoor:


src/specs/missing-css-rules.json
```diff
@@ -66,6 +66,10 @@
     "value": "hsla( <hue> <percentage> <percentage> [ / <alpha-value> ]? )",
     "spec": "https://drafts.csswg.org/css-color/#funcdef-hsla"
   },
+  "<id>": {
+    "value": "<hash-token>",
+    "spec": "https://www.w3.org/TR/2017/CR-css-ui-3-20170302/"
```
tidoust: The definition of `<id>` has been dropped from the Level 3 spec on its way to Recommendation, so I'd rather link to the Editor's Draft, which contains the definition:

`"spec": "https://drafts.csswg.org/css-ui/#typedef-id"`

PR #143, opened 2019-02-28
Exclude `webappsec-` in shortnames from github.io URLs

foolip:
Fixes https://github.com/tidoust/reffy/issues/142.


foolip:
This is untested :)

PR #147, opened 2019-03-22
hack to handle curly quotes in specs

abhishiv:
Hey @tidoust

With reference to tidoust/reffy-reports#31 I found a simple hack to parse curly quotes. Curly quote handling seems to be a bit complicated since they are of two types, so I went with the simples possible solution.

```
  value = value.trim().replace(/‘/g, '\'').replace(/’/g, '\'');
```

Also added a test.

Given it's a hack, I understand if you wouldn't want to merge this and want to implement proper curl quote handling. I'm still making the PR in case anyone else needs it in the meantime.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,1090 @@
+{
+  "name": "reffy",
+  "version": "1.3.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abab": {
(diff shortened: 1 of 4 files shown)


tidoust:
Thanks! I created #148 to track "proper" support for curly quotes in the CSS grammar parser, but let's merge that hack in the meantime.

PR #149, opened 2019-05-08
Add support for CSS range restrictions in CSS Grammar parser

dontcallmedom:
See https://github.com/w3c/csswg-drafts/issues/2921#issuecomment-489494006

### DIFF
--- a/src/lib/css-grammar-parser.js
+++ b/src/lib/css-grammar-parser.js
@@ -37,8 +37,6 @@ const primitives = new Map([
   ["position", {url: "https://drafts.csswg.org/css-values-4/#typedef-position"}]
 ]);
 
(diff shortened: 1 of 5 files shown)


src/lib/css-grammar-parser.js
```diff
@@ -37,6 +37,8 @@ const primitives = new Map([
   ["position", {url: "https://drafts.csswg.org/css-values-4/#typedef-position"}]
 ]);
 
+const numericTypes = ["integer", "number", "percentage", "number-percentage", "length-percentage", "frequency-percentage", "angle-percentage", "time-percentage", "dimension", "length", "angle", "time", "frequency", "resolution"];
```
tidoust: What is this definition for? It does not seem to be used anywhere else in the code.

dontcallmedom: I've now removed it - I had initially thought of throwing when range was used on a non-numeric type, but that goes too close to a validator (vs a parser).

PR #151, opened 2019-06-26
Adapt WebIDL parser to new output of webidl2 parser

dontcallmedom:
Currently, crawl.json doesn't include any WebIDL parsed data because the object model coming out the new version WebIDL2.js is incompatible with the existing code.
This change adapts to some of the changes I detected when running parse-webidl.js on the WebRTC spec

### DIFF
--- a/src/cli/parse-webidl.js
+++ b/src/cli/parse-webidl.js
@@ -144,8 +144,8 @@ function parseIdlAstTree(jsNames, idlNames, idlExtendedNames, externalDependenci
             break;
         case "operation":
             if (def.stringifier) return;
-            parseType(def.idlType, idlNames, externalDependencies, contextName);
-            def.arguments.forEach(a => parseType(a.idlType,  idlNames, externalDependencies, contextName));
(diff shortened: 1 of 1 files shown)


tidoust:
Actually, I do see weird outputs in `crawl.json` such as occurrences of `[object Object]` and JSON structures where simple names are expected but the updates in that PR don't fix them and actually break IDL parsing on my end.

tidoust:
See adjustments I made for parsing of `Global` and `Exposed` in: https://github.com/tidoust/reffy/commit/56dab23a6213a75a77cac6e292dff1c51bb543c0

tidoust:
Hmm, the WebIDL parser fails to parse WebIDL content when Reffy is run on the W3C server. I don't have the same result when I run Reffy locally. Could it be that `npm install` (needed to update webidl2) did not run on the server, @dontcallmedom ?

dontcallmedom:
re-run with a cleaned up checkout - does it produce your expected output?

tidoust:
Yup! When you have time, can you confirm that this pull request is no longer needed?

Issue #134, opened 2018-12-03
Crawl breaks when Respec v23.7.2 is used


tidoust:
Via https://github.com/web-platform-tests/wpt/pull/14333#issuecomment-443723080

That could be due to JSDOM monkeypatch and new "Cannot clone an attribute that does not belong to a document!" error reported during crawl.

tidoust:
Provided #166 gets fixed, we should be able to switch to latest version of Respec when https://github.com/w3c/respec/pull/2187 gets merged.

PR #153, opened 2019-06-27
Add support for spec comment tooling

hiroshige-g:
Hi,

I'm implementing a spec comment tool that matches/updates spec quotes inside C++ source code against latest spec texts, and generates cross references.
https://docs.google.com/document/d/1vmnpbDvMR-ph0v88lC810xsE6kjWVsQP351_l65pxLs/edit

This PR generates spec text data needed for that tool, and generates cross references based on the output of the tool.
(@foolip suggested to integrate this into Reffy to have a single shared scraping infra, at https://chromium-review.googlesource.com/c/chromium/src/+/1570808/12#message-dcd27fceac89d61f789138d8656984741f7dc483)

See
https://docs.google.com/document/d/1vmnpbDvMR-ph0v88lC810xsE6kjWVsQP351_l65pxLs/edit#heading=h.ouxo1agi8os1
for details of data that this PR adds to `crawl.json`. Basically it is spec text, step number, and mapping from the text to DOM locations for each spec step.

What do you think about overall design?

### DIFF
--- a/builds/browser.js
+++ b/builds/browser.js
@@ -789,6 +789,431 @@
     return [...links];
   }
 
+  // Mapping from specDomNodeId to DOM node.
+
+  // Get the next DOM node in the preorder traversal.
+  function nextNode(node) {
+    if (node.firstChild) {
+      return node.firstChild;
+    }
+    while (!node.nextSibling) {
+      node = node.parentNode;
+      if (!node) {
+        return null;
+      }
+    }
+    return node.nextSibling;
+  }
+
+  // Internally, we assign integer IDs to each DOM node, and represent
+  // positions of text segments using pairs of integers
+  // [specDomNodeId, textPositionWithinNode].
+  function assignIDs(node, id) {
+    while (node) {
+      node.specDomNodeId = id;
+      id += 1;
+      node = nextNode(node);
+    }
+  }
+
+  //==============+==============+==============+==============+==============+==============+
+
+  // getRaw* functions returns a list (type SpecStepList) with elements either:
+  // - {text: string, text_map: list of SpecTextMapEntryData},
+  // - {steps: SpecStepList, stepNameStyle: string}, or
+  // - SpecStepList.
+  // See also `SpecTextMapEntryData` in lib/base.py.
+
+  // TODO: Add JSDoc.
+
+  // TODO: refactor by assigning types explicitly. Currently type names are
+  // only in comments for explanation.
+
+  // Returns SpecStepList.
+  // Unlike getRawSteps(), this function only considers textContent and ignores
+  // any DOM structures.
+  function getRawStepsOfText(node) {
+    if (node.nodeType === node.TEXT_NODE) {
+      // Take textContent and remove newlines.
+      const text = node.textContent.replace(/\n/g,' ');
+
+      const text_map = [];
+      let pos = 0;
+      let m;
+      // Split the text into words.
+      while (m = text.substring(pos).match(/[^\s]+/)) {
+        if (m.index > 0) {
+          text_map.push({text: ' '});
+        }
+        const word = m[0];
+        pos += m.index;
+        text_map.push({text: word,
+                       specDomNodeId: node.specDomNodeId,
+                       startTextPosition: pos});
+        pos += word.length;
+      }
+      if (pos < text.length) {
+        text_map.push({text: ' '});
+      }
+
+      return [{text: text, text_map: text_map}];
+    }
+
+    let stepList = [];
+    let c;
+    for(c = node.firstChild; c; c = c.nextSibling) {
+      stepList = stepList.concat(getRawStepsOfText(c));
+    }
+    return stepList;
+  }
+
+  // Returns SpecStepList.
+  function getRawSteps(node) {
+    const window = node.ownerDocument.defaultView;
+    const listStyleType = '';
+        // window.getComputedStyle(node).getPropertyValue('list-style-type');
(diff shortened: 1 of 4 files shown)


foolip:
@dontcallmedom @tidoust have either of you had a chance to look at this?

foolip:
Also @lukebjerring FYI that this is a tool for spec&rarr;impl syncing, relevant for our team to at least be aware of.

tidoust:
Sorry, that took some time. @hiroshige-g this looks pretty interesting. I haven't looked deeply into the code. Some comments on the overall design:

1. This turns `crawl.json` from a big 8.5MB file into a huge 234MB file (including 55MB only for the HTML spec). In theory, that's not a big deal, the file is not meant to be human-readable. That said, for debugging, I often have to go and look into `crawl.json` and was already thinking of splitting it into separate JSON files per spec. That seems even more needed if the results are to include the list of "concepts" that specs define. Would that be a problem for you?
2. The HTML spec remains a challenge. I get "Javascript heap out of memory" from time to time. I suppose that explains why you added `--max-old-space-size=8192`. That's still not very satisfactory. I don't know how to improve the situation though.
3. If I understand things correctly, the cross-reference generation step produces decorated spec files with links to Chromimum source code, using "concepts" in `crawl.json` and spec-source coverage data generated from Chromium's C++ source code. That step seems out of scope for Reffy. In other words, at first glance, I wouldn't integrate the cross reference generator into Reffy (although I do see why it may depend on some of its code for spec retrieval and loading). Or am I missing something?
4. Reffy uses a pinned, monkey-patched (and now outdated) version of ReSpec when it loads an ED, because more recent versions of ReSpec do not run in JSDOM (see #134). This means the final DOM tree that you get from Reffy is not necessarily an exact match of the DOM tree that you get when you load the spec in a browser. How does that affect cross-references?

foolip:
For HTML, scraping the multi-page version might be an option. But I wonder what's taking this much memory. How much memory is required just to parse the single-page spec with JSDOM?

tidoust:
Memory usage of the Node.js process is about `1.1GB` just after parsing the single-page spec with JSDOM (from `80MB` just after that). Processing of the resulting DOM tree adds another `200MB`.

hiroshige-g:
Thanks for feedbacks!

> 1. splitting it into separate JSON files per spec.

Fine with me.

> 2. I get "Javascript heap out of memory" from time to time. I suppose that explains why you added --max-old-space-size=8192.

Yeah, the memory consumption can be problematic in general reffy usage.

(For my usage, the memory consumption wasn't a major problem, because spec comment matching itself is already quite compute-intensive, so adding memory cost wasn't so problemetic for me.)

Does making the concept extraction optional, or moving it into a separate command (i.e. outside whatwg-crawl) alleviate the problem?
(This might drive the concept extraction outside Reffy again though)

> 3. If I understand things correctly, the cross-reference generation step produces decorated spec files with links to Chromimum source code, using "concepts" in crawl.json and spec-source coverage data generated from Chromium's C++ source code.

Yes. I also need the original HTML file itself. So e.g. if reffy saves the original HTML files under `reports` or somewhere and subsequent tools can access these files in addition to `crawl.json`, then that's fine. Does this look a good idea?

> 4. This means the final DOM tree that you get from Reffy is not necessarily an exact match of the DOM tree that you get when you load the spec in a browser. How does that affect cross-references?

This doesn't directly affect the cross-reference generated within this tool, as long as the same DOM is supplied to concept extractor and cross-reference generator.

However, the differences from the DOM on the browser might cause

- Mismatches between the spec texts copy-pasted by Chromium developers from the browser into source codes and the spec texts recognized by the tool, or
- Wrong highlights in the extension-version of cross-reference (because I manage the DOM elements by giving sequential index numbers to each element, and diffs in DOM causes wrong index thus wrong elements to be highlighted). Wrong highlights can anyway occur due to updated spec HTMLs though.

depending on what kind of differences are there.

tidoust:
Hi @hiroshige-g, @foolip. It took a couple of years to upgrade Reffy so that it can better support features like this, but everything should be in place, at last. To create the "spec comments" extracts, all you now need is to:

1. Rename `extract-steps.js` to `extract-steps.mjs` and put it somewhere, for instance in the root folder (it no longer needs to be under `src/browserlib` and there is no `builds` folder anymore)
2. Run the crawler, assuming the module file is in the root folder and that you want extracts to be created under the `reports/test` folder, with:
    ```
    node src/cli/crawl-specs.js --module extract-steps.mjs --output reports/test
    ```

The question that remains is whether this `extract-steps.mjs` module should be integrated in Reffy as a core processing module, in other words whether you'd like these steps extracts to appear in Webref. We don't really expect to do anything useful with that data in the near future. Is this still current and useful on your side?

The extraction produces `~380MB` of data (the HTML spec being the largest export with `~50MB`). By comparison, the `ed` folder in Webref currently weighs `<80MB`, so we're talking about a `x6` increase in data. That's not necessarily a problem, especially now that we release dedicated NPM packages of Webref. I just want to make sure that there exists some sort of feedback loop on the data that gets produced.

tidoust:
I'll take lack of response as consent and close this PR ;-)

Additional notes:
- the crawler has become the main CLI tool in latest version of Reffy, meaning that the suggested command above would become:
    ```bash
    node reffy.js --module extract-steps.mjs --output reports/test
    ```
- FWIW, I made a short video to [demo Reffy's CLI tool](https://www.w3.org/2021/10/TPAC/demos/reffy.html) for TPAC 2021.

PR #154, opened 2019-07-23
Add Largest Contentful Paint to specs list

foolip:


foolip:
@tidoust @dontcallmedom, https://github.com/tidoust/reffy-reports/commit/66f78e57928fb1d5d384d65731149dccc77b3110 was the first reffy-reports commit after this and https://github.com/tidoust/reffy/pull/155 were merged, and they do mention the two added specs, but the IDL files are not created. This is despite the parsed IDL showing up in whatwg/crawl.json.

Is some manual work needed to add the files the first time?

tidoust:
Yes, now done!

foolip:
@tidoust is there any way that step could happen automatically? I'd like to smooth out the pipeline between spec creation and idlharness tests existing as much as possible :)

tidoust:
@foolip Technically, that's doable. I remember @dontcallmedom argued against it in the past. One problem is that, given crawls still often fail on several specs, we probably want to make sure first that the crawl can never produce badly named IDL files (e.g. `undefined.idl` or `spec-title-not-found.idl`) that should not exist.

foolip:
I see, there are some mishaps you'd like to prevent, just as in our downstream use of reffy-reports in web-platform-tests. There we create PRs which humans review and merge. Perhaps that would be a middle ground here, but it's still a bit of work to set that up, perhaps not worth it to automate a once-per-spec action.

PR #157, opened 2019-08-28
Run crawl on GitHub actions (results pushed to reffy-reports)

foolip:
Fixes https://github.com/tidoust/reffy/issues/146.

### DIFF
--- a/.github/workflows/crawl.yml
+++ b/.github/workflows/crawl.yml
@@ -1,0 +1,44 @@
+on:
+  schedule:
+    - cron: '30 */6 * * *'
+  push:
+    branches:
+    - master
+name: crawl
+jobs:
+  crawl:
+    runs-on: ubuntu-18.04
+    steps:
+    - name: checkout reffy
+      uses: actions/checkout@master
+    - name: setup node
+      uses: actions/setup-node@master
+      with:
+        node-version: '10.x'
+    - name: install pandoc
+      run: sudo apt-get install pandoc
+    - name: setup reffy
+      run: |
+        echo "${{ secrets.CONFIG_JSON }}" | base64 --decode > config.json
+        npm ci
+    - name: reffy crawl
+      run: npm run all
+    - name: checkout reffy-reports
+      uses: actions/checkout@master
+      with:
+        repository: tidoust/reffy-reports
+        ref: master
+        path: reffy-reports
+    - name: copy reports
+      run: rsync -av --exclude=README.md reports/ ../reffy-reports/
+    - name: push reffy-reports
+      working-directory: ../reffy-reports
+      run: |
+        git config user.email auto@foolip.org
+        git config user.name autofoolip
+        git remote set-url --push origin https://autofoolip:${{ secrets.REFFY_REPORTS_TOKEN }}@github.com/foolip/reffy-reports.git
+        git add -A
+        git commit -m 'Update from new reffy run'
+        #npm version patch
+        git push origin HEAD:master
+        #git push origin --tags
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,1090 @@
+{
+  "name": "reffy",
+  "version": "1.3.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abab": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/abab/-/abab-2.0.0.tgz",
+      "integrity": "sha512-sY5AXXVZv4Y1VACTtR11UJCPHHudgY5i26Qj5TypE6DKlIApbwb5uqhXcJ5UUGbvZNRh7EeIoW+LrJumBsKp7w=="
+    },
+    "acorn": {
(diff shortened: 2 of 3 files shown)


foolip:
https://github.com/foolip/reffy/runs/205624540 shows this in action. It's basically working, but to actually get it set up on the reffy repo and turn down the existing crawl infrastructure we'll have to coordinate.

.github/workflows/crawl.yml
```diff
@@ -0,0 +1,44 @@
...
+        ref: master
+        path: reffy-reports
+    - name: copy reports
+      run: rsync -av --exclude=README.md reports/ ../reffy-reports/
+    - name: push reffy-reports
+      working-directory: ../reffy-reports
+      run: |
+        git config user.email auto@foolip.org
+        git config user.name autofoolip
+        git remote set-url --push origin https://autofoolip:${{ secrets.REFFY_REPORTS_TOKEN }}@github.com/foolip/reffy-reports.git
```
foolip: An alternative to this would be to put the cron job in reffy-reports instead, and have that job clone/install reffy. That would avoid the need for a separate token. @tidoust do you have a preference for where the updating infra lives?

foolip: Tied in with that is this question: should we do runs with the master branch of reffy, or `npm install reffy` to use the most recent published version?

foolip:
Ping @tidoust. Have you opted in to the GitHub Actions beta? We found in https://github.com/plinss/widlparser/pull/48 that you don't get it enabled immediately, so if you're not already using it, this setup won't actually work in this repo right away.

If you do want to put the crawl on a CI system like this, I'd be happy to put it on Travis instead, or Azure Pipelines which has a cron feature very similar to GitHub Actions. (GitHub Actions appears to be built on the same underlying infrastructure as GitHub Actions, both run on Azure.)

.github/workflows/crawl.yml
tidoust: I think I'd prefer to put the logic in reffy-reports and leave the reffy repo as code-only, with reffy-reports containing the logic to generate the reports, the reports themselves, and the logic to update WPT.

Given that I haven't been consistent about updating the published version of Reffy on npm and that I don't see any good reason to incorporate changes in the master branch that we wouldn't want to see reflected in the next crawl, it may be better to rely on the master branch. I don't feel strongly either way.

tidoust:
I had not opted in to the GitHub Actions beta. Just did that and I have indeed been put on a waitlist.

I'm personally not married to any particular system. What we need is:
- A system with enough RAM
- A system where we can run jobs at least once per day
- Ideally, a system where we can also run the job manually once in a while, e.g. to re-run it if it failed, or to update the report after we make changes to Reffy's code
- A system that can give us a log report about what happened
... with "we" being all parties interested in reports (you, other WPT folks, @dontcallmedom, me)

tidoust:
Code looks good otherwise, thanks for preparing it!
I note the `git add -A` that may end up adding weird IDL files if crawl runs into issues as discussed previously. It won't remove existing files if I understand the `rsync` command properly, so not a real problem. Also see https://github.com/tidoust/reffy-reports/issues/12

.github/workflows/crawl.yml
foolip: Alright, putting it in reffy-reports would avoid the need for `REFFY_REPORTS_TOKEN`, I can move to there and also figure out how to trigger the wpt sync on each update. (With GitHub Actions, a push done by one action can't trigger another action, so they'll need to happen together.)

foolip:
@tidoust we could wait until the GitHub Actions invite comes through, but do you already use Travis, Azure Pipelines or any other CI system on any of your repos? Converting to use any CI system you're already signed up for shouldn't be much work.

The amount of memory required is the only thing that seems potentially problematic, but the CI systems I've mentioned all have 6-7.5GB of RAM available:
- https://help.github.com/en/articles/virtual-environments-for-github-actions#supported-virtual-environments-and-hardware-resources  (7 GB)
- https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system (7.5 GB)
- https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#capabilities-and-limitations (6 GB of physical memory and 13 GB of total memory)

tidoust:
Travis is enabled on `reffy-reports`. I've never used other CI systems for now.

Another limit we may hit is that Reffy generates a lot of outputs on the console (most of it comes from ReSpec execution and is essentially garbage for us, but it's hard to filter out noise from potentially useful errors).

CI systems may restrict log size. For instance, it is 4MB on Travis:
https://docs.travis-ci.com/user/common-build-problems/#log-length-exceeded
(I haven't checked whether that's enough)

tidoust:
PR to be moved (and adjusted) to reffy-reports. Closing this PR without merging as a result.

@foolip, note I now see an "Actions" tab on my repos!

foolip:
Excellent, I guess that means you're in the beta. I'll send the PR for reffy-reports now.

PR #160, opened 2019-09-09
Put package-lock.json under source control

foolip:
Part of https://github.com/tidoust/reffy/issues/159.


foolip:
I split this out from https://github.com/tidoust/reffy/pull/157. When installing reffy in CI it's a good idea to `npm ci`, and that requires package-lock.json to be present.

PR #177, opened 2019-09-24
Update animation-worklet/scroll-animations URL

saschanaz:
This is manual update, but is there an automatic way to do this?

Closes https://github.com/saschanaz/webidl-updater/issues/4


tidoust:
Thanks @saschanaz, I had started to prepare a more complete update for CSS specs as some of them were missing from the crawl. Now done in PR #179, which includes the changes you're suggesting (although with the /TR/ URL for the CSS Animation Worklet spec). I'll close this PR without merging and merge #179 instead.

Update remains manual. There's a small spec finder tool in Reffy (see [Other tools](https://github.com/tidoust/reffy/#other-tools)) which can be used to detect missing specs, and which I used to prepare the PR, but it needs a bit of love and returns a few false positives for now.

Issue #166, opened 2019-09-09
Adapt code to run with latest version of JSDOM


tidoust:
See https://github.com/tidoust/reffy/pull/163#issuecomment-529494042

PR #180, opened 2019-09-26
Adapt code to run with latest version of JSDOM

tidoust:
JSDOM now exposes a new ResourceLoader class close to the one we were using to intercept HTTP requests. This update switches to the new version, and drops a couple of patches that are no longer required (most of them are still needed though).


tidoust:
The PR fixes #166 and replaces #163.

PR #193, opened 2019-11-13
Drop Pointer Events Extensions from spec list

foolip:
Removed in https://github.com/w3c/pointerevents/pull/310.


foolip:
I've also sent https://github.com/tidoust/reffy-reports/pull/43.

PR #195, opened 2019-11-21
Get back to RespecDocWriter to generate specs

### LINKED ISSUE #134: Crawl breaks when Respec v23.7.2 is used

tidoust:
Making latest version of Respec run under JSDOM is too much of a maintenance burden in the end:
https://github.com/w3c/respec/issues/1469#issuecomment-546790835

This update makes Reffy use respecDocWriter (which uses the headless browser Puppeteer under the hoods) again. This allows to run the latest version of Respec.

JSDOM is still used to load specifications that do not use Respec and to detect when a specification uses Respec. Note the code no longer runs external scripts when it loads specifications: that seems unneeded in practice.

The code still contains minimal monkey patching for JSDOM so because a number of specifications do have inline scripts that call `window.matchMedia`, which JSDOM does not support.

This will fix #134.

Note I ran local crawls to check the results. This triggers a number of updates to the list of references that get extracted, but they all seem correct (and long overdue). It triggers some formatting updates to IDL extracts but no substantive changes. Only current issue is that some specs are now flagged by Respec as purely informative because they lack a conformance section and Reffy thus fails to extract WebIDL content for now for these specs, see discussion in:
https://github.com/w3c/respec/issues/2580
(I prepared PR against these specs)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,10 +4,172 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@snyk/cli-interface": {
+      "version": "2.3.0",
+      "resolved": "https://registry.npmjs.org/@snyk/cli-interface/-/cli-interface-2.3.0.tgz",
+      "integrity": "sha512-ecbylK5Ol2ySb/WbfPj0s0GuLQR+KWKFzUgVaoNHaSoN6371qRWwf2uVr+hPUP4gXqCai21Ug/RDArfOhlPwrQ==",
+      "requires": {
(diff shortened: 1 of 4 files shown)


tidoust:
@foolip FYI, this move to the latest version of Respec will trigger a few updates to IDL exports because Respec formats WebIDL snippets slightly differently to ease readability.

PR #208, opened 2020-01-03
Remove any .idlHeader descendents in extractRespecIdl

foolip:
Adapting to https://github.com/w3c/respec/pull/2663.

Fixes https://github.com/tidoust/reffy/issues/205.

### DIFF
--- a/src/cli/extract-webidl.js
+++ b/src/cli/extract-webidl.js
@@ -158,9 +158,13 @@ function extractRespecIdl(doc) {
             .filter(el => !el.closest(nonNormativeSelector))
             .map(el => el.cloneNode(true))
             .map(el => {
+                const header = el.querySelector('.idlHeader');
+                if (header) {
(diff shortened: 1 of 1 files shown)


foolip:
@tidoust I can't quite figure out the testing setup for this repo. Both jasmine and mocha are dependencies and I see some tests in spec/*. `npx jasmine` does find and try to run these, but there are failures.

I see https://github.com/tidoust/reffy/issues/165 but what should I do with a change like this?

sidvishnoi:
@foolip There will be another markup change in ReSpec WebIDL blocks (https://github.com/w3c/respec/pull/2701).
It won't break reffy, but letting you know as you asked :slightly_smiling_face:

foolip:
Thanks @sidvishnoi! Happily I didn't select for the element type, so you're right this shouldn't affect Reffy.

PR #209, opened 2020-01-03
Crawl specs in same process for easier debugging

foolip:


foolip:
@tidoust this is what I did to be able to debug https://github.com/tidoust/reffy/issues/205. Would you like to have something like this checked in, or is it not worth it? The mode will probably break if it's not used by any tests...

tidoust:
Thanks @foolip! I just pushed an update that incorporates your code to create a debug mode for Reffy, see [commit message](https://github.com/tidoust/reffy/commit/651057de737b90eea927e24687bec3c490bf6ae4) for details. Note that I also made the crawling sequential in that mode to further ease debugging.

foolip:
Thanks @tidoust, those changes look good, and I'll try it the next time I need to debug!

PR #212, opened 2020-01-03
Remove unused respec dependency

foolip:
`require('respec')` is nowhere to be found, and the appearance of
pinning respec was misleading in
https://github.com/tidoust/reffy/issues/205.

It appears that respec will simply be loaded from whatever the spec
points to, such as https://www.w3.org/Tools/respec/respec-w3c.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,172 +4,10 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@snyk/cli-interface": {
-      "version": "2.3.0",
-      "resolved": "https://registry.npmjs.org/@snyk/cli-interface/-/cli-interface-2.3.0.tgz",
-      "integrity": "sha512-ecbylK5Ol2ySb/WbfPj0s0GuLQR+KWKFzUgVaoNHaSoN6371qRWwf2uVr+hPUP4gXqCai21Ug/RDArfOhlPwrQ==",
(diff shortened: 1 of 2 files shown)


tidoust:
Wait, there **is** a [`require` statement on respec](https://github.com/tidoust/reffy/blob/master/src/lib/util.js#L11).

The treatment of Respec specs is a two-step process: Reffy first loads the spec as a regular HTML spec (simply because it doesn't know yet whether the document is using Respec), and then [uses `respecDocWriter` to generate the spec](https://github.com/tidoust/reffy/blob/master/src/lib/util.js#L130), so the dependency is needed (or we'd need to duplicate the code in `respecDocWriter` to load the page through Puppeteer).

What is indeed confusing is that `respecDocWriter` does not pin the version of Respec used in the spec to the version of Respec it comes bundled with.

foolip:
Oops, despite efforts to grep a bit more broadly I failed to spot that.

So it looks like only https://github.com/w3c/respec/blob/develop/tools/respecDocWriter.js is used, and that's a utility that just loads a page in Puppeteer and tries to work with whatever version of ReSpec happens to be running within.

https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagesetrequestinterceptionvalue could probably be used to pin the ReSpec version fetched in a somewhat hacky way.

PR #222, opened 2020-01-20
Move MIME Sniffing from blacklist.json to specs-other.json

foolip:
Its definitions are used a lot in HTML:
https://html.spec.whatwg.org/#mime-type


foolip:
The background here beyond getting the lists in closer sync is that I would like Reffy to extract a list of definitions, since that could potentially replace Shepherd (with its own spec list) for Bikeshed's cross-spec referencing. Do you think that would be OK?

tidoust:
Yes, that's something I have in mind as well. I had started to work on extraction of exported terms in a separater branch. I just turned that into a [draft pull request](https://github.com/tidoust/reffy/pull/229) to discuss needs and track progress.

PR #224, opened 2020-01-20
Add more specs to specs-other.json

foolip:
These are specs present in https://foolip.github.io/day-to-day/ but not
in Reffy.

### DIFF
--- a/src/specs/specs-other.json
+++ b/src/specs/specs-other.json
@@ -1,8 +1,24 @@
 [
   "https://streams.spec.whatwg.org/",
+  "https://w3c.github.io/accname/",
+  "https://w3c.github.io/core-aam/",
   "https://w3c.github.io/editing/contentEditable.html",
+  "https://w3c.github.io/editing/execCommand.html",
+  "https://webassembly.github.io/spec/",
+  "https://wicg.github.io/origin-policy/",
+  "https://wicg.github.io/priority-hints/",
+  "https://wicg.github.io/web-share-target/",
+  "https://www.w3.org/TR/clear-site-data/",
   "https://www.w3.org/TR/mixed-content/",
+  "https://www.w3.org/TR/network-error-logging/",
+  "https://www.w3.org/TR/payment-method-id/",
+  "https://www.w3.org/TR/payment-method-manifest/",
   "https://www.w3.org/TR/preload/",
+  "https://www.w3.org/TR/resource-hints/",
+  "https://www.w3.org/TR/timing-entrytypes-registry/",
+  "https://www.w3.org/TR/trace-context/",
   "https://www.w3.org/TR/uievents-code/",
-  "https://www.w3.org/TR/uievents-key/"
+  "https://www.w3.org/TR/uievents-key/",
+  "https://www.w3.org/TR/upgrade-insecure-requests/",
+  "https://www.w3.org/TR/webrtc-svc/"
 ]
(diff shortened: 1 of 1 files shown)


src/specs/specs-other.json
```diff
@@ -1,8 +1,24 @@
 [
   "https://streams.spec.whatwg.org/",
+  "https://w3c.github.io/accname/",
+  "https://w3c.github.io/core-aam/",
   "https://w3c.github.io/editing/contentEditable.html",
+  "https://w3c.github.io/editing/execCommand.html",
+  "https://webassembly.github.io/spec/",
```
foolip: @tidoust I see you've excluded TC39 specs. Do you want to exclude this too?

src/specs/specs-other.json
```diff
@@ -1,8 +1,24 @@
 [
   "https://streams.spec.whatwg.org/",
+  "https://w3c.github.io/accname/",
+  "https://w3c.github.io/core-aam/",
   "https://w3c.github.io/editing/contentEditable.html",
+  "https://w3c.github.io/editing/execCommand.html",
```
foolip: Not sure this should be added. @tidoust what do you want to do for specs that aren't implemented yet or that are abandoned? Add them to blacklist.json?

tidoust: I would put specs that aren't implemented yet in the list of specs crawled, and add specs that are abandoned to blacklist.json... unless they are implemented and on a long transition path to being abandoned perhaps.

In this particular case, I think I'd leave it out since we have planned replacements (contentEditable and input events) in the list. But if you feel that's useful, fine to have it in!

src/specs/specs-other.json
foolip: This was because I hadn't updated the manual fixups for day-to-day. I've done so in https://github.com/foolip/day-to-day/commit/f01e6dc78faa8ba0ea64c87a9c44ebf1d5994625 and will revert this change.

Crawling https://webassembly.github.io/spec/core/ might be nice, but it's multi-page and would probably need some special treatment.

src/specs/specs-other.json
foolip: I don't think `execCommand` is ever going to be removed, but nobody has any appetite for working on the spec, so I'll drop this. Might add it if later some reference to it is unresolved.

src/specs/specs-other.json
```diff
@@ -1,8 +1,24 @@
...
+  "https://www.w3.org/TR/payment-method-manifest/",
   "https://www.w3.org/TR/preload/",
+  "https://www.w3.org/TR/resource-hints/",
+  "https://www.w3.org/TR/timing-entrytypes-registry/",
+  "https://www.w3.org/TR/trace-context/",
   "https://www.w3.org/TR/uievents-code/",
-  "https://www.w3.org/TR/uievents-key/"
+  "https://www.w3.org/TR/uievents-key/",
+  "https://www.w3.org/TR/upgrade-insecure-requests/",
+  "https://www.w3.org/TR/webrtc-svc/"
```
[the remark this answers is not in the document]
foolip: Good catch!

src/specs/specs-other.json
tidoust: I don't know if all WebAssembly variants are kept in sync automatically, but adding https://www.w3.org/TR/wasm-core-1/ would mean that the corresponding Editor's Draft at https://webassembly.github.io/spec/core/bikeshed/ gets crawled, and that one is single page.

tidoust:
PR looks good to me. Are you planning further changes, @foolip?

foolip:
@tidoust this is good to go, no further changes.

PR #226, opened 2020-01-21
Add SVG Animations spec

foolip:


foolip:
Turns out the IDL in this spec is broken, I've filed https://github.com/w3c/svgwg/issues/775.

PR #229, opened 2020-01-21
Extract terms defined with "data-export"

tidoust:
First stage towards extracting definitions in a spec. Current code just extracts terms flagged with a `data-export` attribute that can be referenced from other specifications.

Not yet ready to merge.


foolip:
Extracting definitions using Reffy is something I recently discussed with @tabatkins.

To provide the data for cross-spec linking in Bikeshed, it would have be possible to generate the files in https://github.com/tabatkins/bikeshed-data/tree/master/data/anchors from this source, and https://github.com/tabatkins/bikeshed/blob/6a4c097ae8b731ee02752a258e8ac5ffbb752299/bikeshed/update/updateCrossRefs.py#L313 is the code that generates those files.

It will be important to extract the type of the definition, and these are documented in https://tabatkins.github.io/bikeshed/#dfn-types.

tabatkins:
Note that Bikeshed needs *all* terms; it treats exported terms a little differently from unexported, but does still allow linking to unexported terms if you're explicit about that being what you want.

tabatkins:
> It will be important to extract the type of the definition

This'll be present in data-dfn-type attributes on the `dfn` or `h1-6`.

To be specific, <https://tabatkins.github.io/bikeshed/#dfn-contract> defines what Bikeshedded documents will expose (and what other specs need to expose to interop with the Bikeshed ecosystem). That is:

* dfns, and potentially headings, can be definitions
* they *must* have an id attribute
* they *must* have a data-dfn-type attribute
* they *may* have a data-export or data-noexport boolean attribute
* they *may* have a data-dfn-for attribute
* they will either have a data-lt attribute giving the linking text, or else use their textContent (note that an empty `data-lt` **is significant** - it indicates the `dfn` isn't a link target at all; only the lack of a `data-lt` indicates that you should use the text content)

tidoust:
PR still needs to be finalized, but code should now follow the definitions data model.

@tabatkins, I note that the rule for namespaces is slightly problematic. It says "These definitions MUST contain a data-dfn-for attribute, containing a comma-separated list of one or more definitions they’re namespaced to" but, in practice, namespace names may have commas which don't seem to be escaped in any way, so splitting on "," is not going to always work.

For instance, if I look at the definition of the [`handler` parameter in Media Session](https://w3c.github.io/mediasession/#dom-mediasession-setactionhandler-action-handler-handler), its `data-dfn-for` attribute is `MediaSession/setActionHandler(action, handler)` and the comma there is not meant to be a separator between namespace names.

That seems to only happen for IDL functions. I used [a small regexp to trap that](https://github.com/tidoust/reffy/pull/229/files#diff-523f0e374d4eca971d65363c11cea711R60). Question is: is it enough?

More generically, does the [extraction function](https://github.com/tidoust/reffy/pull/229/files#diff-523f0e374d4eca971d65363c11cea711R1) (which would run on the document) look good?

It would produce results such as:

```json
{
  "id": "event-abort",
  "href": "https://html.spec.whatwg.org/multipage/indices.html#event-abort",
  "linkingText": [
    "abort"
  ],
  "localLinkingText": [],
  "type": "event",
  "for": [
    "Window"
  ],
  "exported": true
}
```

tabatkins:
Yes, technically the rule is more complicated to allow for commas inside of parens, and I should update the docs to reflect that. I'm using the following to split apart for values:

```python
[value.strip() for value in re.split(r',(?![^()]*\))', forValues)
```

This looks extremely similar to yours, so we should be good.

foolip:
@tidoust is this at a point where you want review yet?

tidoust:
> @tidoust is this at a point where you want review yet?

We're close but not yet ;)

Multi-page processing does not work that well for now. At least, code seems to generate different IDL results for HTML and SVG2 when it parses the single-page version or when it parses pages one by one. I need to look into it.

I would like to add a normative/informative flag as well. I don't know if Bikeshed needs that, but that seems useful to spot potential cross-spec hiccups in any case.

tidoust:
> Multi-page processing does not work that well for now. At least, code seems to generate different IDL results for HTML and SVG2 when it parses the single-page version or when it parses pages one by one. I need to look into it.

Well, I know why, actually. Current code assumes that there will be only page per section, but there are 2-3 pages per section in some cases. Page detection logic needs to be improved.

tabatkins:
> I would like to add a normative/informative flag as well. I don't know if Bikeshed needs that, but that seems useful to spot potential cross-spec hiccups in any case.

Not currently used, no. Bikeshed just yells at you if you try to define something in an informative section. (When it can detect that.)  But if you're detecting that, then great, I can probably use the info.

tidoust:
OK, we should now be good enough to go.

Some notes:
- The update will change the URL of the HTML (resp. SVG) spec that appears in the IDL extracts to `https://html.spec.whatwg.org/multipage/` (resp. `https://svgwg.org/svg2-draft/`).
- Some IDL interface definitions in HTML have `// also has obsolete members` in the single page version, but not in the multipage version. For instance, see the html element: [single page](https://html.spec.whatwg.org/#htmlhtmlelement), [multipage](https://html.spec.whatwg.org/multipage/semantics.html#htmlhtmlelement). Since code now crawls the multipage version, these comments will disappear from the IDL extract.
- Definitions extraction generates a JSON extract per spec in a `dfns` subfolder.
- Definitions will also appear in `crawl.json`, which means the size of the file jumps to ~20MB.
- Definitions descriptions contain two boolean properties: `exported` and `informative`. I haven't tried to turn them into enums, but I happily take suggestions if people feel strongly about that.

Things that may need adjustments later on:
- Logic is slightly inconsistent because it creates one CSS/IDL extracts per shortname (without version), but definitions are for each and every specs.
- The `getShortname` function's behavior is a bit weird, given that it trips versions from TR shortnames but not from other names. That will get fixed when we switch to browser-specs.

src/lib/util.js
```diff
@@ -221,54 +221,53 @@ async function processSpecification(spec, callback, args, counter) {
...
+            'https://svgwg.org/svg2-draft/': '#toc a[href]',
         };
         if (multiPagesRules[page.url()]) {
-            let urls = await page.$$eval(multiPagesRules[page.url()], links =>
-                links.map(link => (new URL(link.getAttribute('href'), link.ownerDocument.baseURI)).toString()));
+            // Extract URLs that appear in the TOC and drop duplicates
+            const tocUrls = await page.$$eval(multiPagesRules[page.url()],
+                links => links.map(link =>
+                    (new URL(link.getAttribute('href'), link.ownerDocument.baseURI)).toString()));
+            const pageUrls = tocUrls
```
dontcallmedom: maybe use a set to convey more directly the uniqueness semantics (but I don't feel strongly at all about it)

dontcallmedom:
great job, thanks!

src/lib/util.js
tidoust: True. Done!

tidoust:
Thanks @dontcallmedom, I also switched the `exported` boolean flag to an `access` enum (with values `public`, and `private`) to "prepare for the great future".

tidoust:
@marcoscaceres This PR would make Reffy extract definitions from specs (to build a cross-ref database). I believe that Respec now follows Bikeshed's definitions data model. Are there specificities that we should handle though?

In particular, is the code in [src/browserlib/extractdfns.js](https://github.com/tidoust/reffy/pull/229/files#diff-523f0e374d4eca971d65363c11cea711R1), run against a loaded HTML spec, ok for specs generated with Respec too? (That seems to be the case, but we may have missed corner cases)

marcoscaceres:
>  I believe that Respec now follows Bikeshed's definitions data model. Are there specificities that we should handle though?

No. If they don't align, that's a bug and feel free to yell at me and @sidvishnoi about it :) 

> In particular, is the code in src/browserlib/extractdfns.js, run against a loaded HTML spec, ok for specs generated with Respec too? (That seems to be the case, but we may have missed corner cases)

It should be fine*. Would you like us to review the JS? If yes, I or Sid can set some time aside to do that. 

*terms and conditions apply.

tidoust:
> It should be fine

Great! I have of course not reviewed the terms and conditions but what could go wrong? :)


> Would you like us to review the JS? If yes, I or Sid can set some time aside to do that.

Not necessarily. You're certainly welcome to review the code but nothing depends on it for now, so I think it's fine to just put it in place and iterate from there. We'll have opportunities to assert whether the list of definitions produced looks good (e.g. by comparing the results with Shepherd data for specs that are in both systems).

What I'm more interested to know is whether the information for each definition would be enough to cover possible needs. But that too can be done later on, if and when you start consuming that data.

marcoscaceres:
Ok, cool. So yeah, give us a shout if you need anything. Stay healthy! thinking of you all.

PR #231, opened 2020-01-21
Treat all specs the same by not expecting CSS/IDL up front

foolip:
This is in preparation for using a flat list of specs from
https://github.com/w3c/browser-specs. This by necessity removes some
useful checks that depend on the human categorization into
css/idl/other.


foolip:
@tidoust I'm leaving this as a draft as I need feedback from you. I think this works, but how to vet it? Do you just do `npm run ed-crawl` once with and once without changes and compare the output to see if it's OK?

How bad is it that part of the error reporting goes missing with this?

This seems like a necessary step to use a single spec list as input, but if there are other ways please suggest!

src/cli/study-crawl.js
```diff
@@ -161,23 +160,13 @@ function studyCrawlResults(results, specsToInclude) {
                 // Whether the spec normatively references the WebIDL spec
                 // (all specs that define IDL content should)
                 noRefToWebIDL: (spec !== WebIDLSpec) &&
-                    spec.flags.idl &&
+                    // TODO: doesn't work because `saveIdl` deletes `spec.idl.idl`.
+                    spec.idl.idl &&
```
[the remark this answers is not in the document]
tidoust: Actually, that should rather be `(spec.idl.bareMessage || (idlDfns.length > 0) || (idlExtendedDfns.length > 0))`, given that `message` no longer is the right property to check to detect the presence of an error (see https://github.com/tidoust/reffy/commit/234ee19b6f39c89633d927c819edbf7497aee9f1)

tidoust:
@foolip I adjusted the pull request to:

1. Integrate feedback I suggested
2. Flag delta specs as such in the list of specs with a "delta" keyword.
3. Add a couple of CSS delta specs that were in Shepherd but not in Reffy
4. Drop duplicates in CSS and IDL spec lists
5. Fix crawling of the CSS 2.2 draft (the TOC format changes a few days ago)
6. Handle delta specs. The code produces CSS and IDL extracts for the latest level of a spec that is not a delta spec for now. The only practical change that this introduces on the IDL extracts is for CSS Images because [CSS Images Level 4](https://drafts.csswg.org/css-images-4/) is the one that defines some IDL content but it's also a delta spec, and [CSS Images Level 3](https://drafts.csswg.org/css-images-3/) does not have any IDL. Except for CSS Images, the crawl generates the exact same IDL extracts with the new code. There are more changes in the CSS extracts because they used to be the ones of the latest level, and in practice of the delta spec if one existed, whereas they become the CSS extracts of the latest level that is not a delta spec with the new code. That seems a more useful result.

Is the CSS Images case a big deal? If so, one alternative would be to amend the rule to "Create the IDL extract for the latest full level that defines some IDL, or to the latest delta level if that's the only one that has IDL definitions", that just seems slightly less logical. Another alternative would be to produce CSS/IDL extracts for delta specs named `shortname-delta.json` and `shortname-delta.idl`, but these files would be temporary by definition (delta specs should eventually all become full specs), and I'm not sure you'd want to see such temporary files in Web Platform Tests (although one could argue that "temporary" in standardization can sometimes mean "for a very long time"...)

foolip:
Thanks for the detailed explanation of the changes, @tidoust!

Including https://drafts.csswg.org/css-images-4/#idl-index seems like a good thing. Even if that isn't implemented anywhere, that's the case for a bunch of other things as well, and having the IDL extract is better than not having it. It's really only IDL that we think is never going to be implemented but still left in some spec that could turn into a nuisance.

Using "latest level of a spec that is not a delta spec for now" sounds like a good start, and if we want to handle delta specs on top of that, that seems fine to punt until later. If there's no overlap then just extracting all CSS and IDL for delta specs would work. But later :)

tidoust:
Actually, the update introduces another IDL change: it creates a new `svg-markers.idl` file for [SVG Markers](https://svgwg.org/specs/markers/#idl), which wasn't listed as an IDL spec before.

src/cli/crawl-specs.js
```diff
@@ -368,45 +370,21 @@ async function saveResults(crawlInfo, crawlOptions, data, folder) {
...
-        .filter(spec => isLatestLevel(spec, 'idl'))
+        .filter(spec => !spec.flags.delta)
+        .filter(spec => spec.idl && spec.idl.idl)
+        .filter(spec => isLatestLevel(spec, data))
         .map(saveIdl));
 
     // Save CSS dumps for the latest level of a spec to the css folder
     await Promise.all(data
-        .filter(spec => spec.flags.css && spec.css && (
+        .filter(spec => !spec.flags.delta)
```
foolip: `isLatestLevel` is documented as "The function ignores delta specs, which are never considered to be the last level" so could the delta check be dropped here and elsewhere?

src/cli/study-crawl.js
```diff
@@ -94,9 +95,11 @@ function studyCrawlResults(results, specsToInclude) {
         .filter(uniqueFilter);
     var idlNamesIndex = {};
     knownIdlNames.forEach(name => {
-        idlNamesIndex[name] = results.filter(spec => {
+        let list = results.filter(spec => {
             return spec.idl && spec.idl.idlNames && spec.idl.idlNames[name];
         });
+        list = list.filter(spec => isLatestLevel(spec, list));
```
foolip: How about folding this into the above filter, or at least chaining them to get `const list`?

src/lib/util.js
```diff
@@ -497,11 +497,50 @@ function getShortname(spec) {
...
+ * follow the old `css3-` pattern, but we're only interested in comparing
+ * with more recent levels in that case, so it does not matter)
+ *
+ * The function ignores delta specs, which are never considered to be the last
+ * level.
+ *
+ * @function
+ * @public
+ * @param {Object} spec Spec to check
+ * @param {Array(Object)} list List of specs (may include the spec to check)
```
foolip: It looks like `list` has to include the spec to check for the code to work, so not "may include".

src/lib/util.js
```diff
@@ -497,11 +497,50 @@ function getShortname(spec) {
...
+ * @return {Boolean} true if the spec is the latest level to consider.
+ */
+function isLatestLevel(spec, list) {
```
foolip: This looks like it would be pretty easy to add tests for in tests/util.js, if you'd like.

src/specs/specs-css.json
```diff
@@ -1,52 +1,59 @@
...
+  "https://drafts.csswg.org/css-backgrounds-4/ delta",
   "https://drafts.csswg.org/css-box-3/",
-  "https://drafts.csswg.org/css-color-5/",
-  "https://drafts.csswg.org/css-conditional-4/",
+  "https://drafts.csswg.org/css-conditional-4/ delta",
   "https://drafts.csswg.org/css-env-1/",
   "https://drafts.csswg.org/css-exclusions-1/",
-  "https://drafts.csswg.org/css-gcpm-4/",
-  "https://drafts.csswg.org/css-multicol-2/",
+  "https://drafts.csswg.org/css-gcpm-4/ delta",
```
foolip: Is the "is delta" bit something we'll need to maintain in https://github.com/w3c/browser-specs/? If so can you comment on https://github.com/w3c/browser-specs/issues/1?

foolip:
That sounds good. As long as no IDL is unexpectedly deleted there's no much damage that can be done. Did you delete all files first so that deletions would be detected?

src/cli/crawl-specs.js
tidoust: I'll update the documentation. I didn't mean "ignores" as "does not handle" but rather as "makes sure that delta specs are not considered when looking for a newer level".

src/cli/study-crawl.js
tidoust: That's not the cleanest code for sure but I cannot chain directly: `isLatestLevel` needs to be called with the already filtered list of specs that define the term to handle cases where an IDL term is defined in level X and is no longer defined in level X+1 (that should not happen often in practice. Actually, that's something that could be worth reporting as a potential anomaly).

src/lib/util.js
tidoust: Oops, indeed!

src/specs/specs-css.json
tidoust: Yes. As far as I can tell, this is what we need to track the notion of "current level" that was raised in w3c/browser-specs#5. That is, all specs in Shepherd for which the "current level" is not the latest level are delta specs in practice. [`css-break-4`](https://drafts.csswg.org/css-break-4/) seems to be one exception to the rule, and I don't quite understand which it's not considered to be the "current level". Anyway, I'll comment over there.

src/lib/util.js
tidoust: Oops, indeed!

src/lib/util.js
tidoust: Reaction was for previous comment actually. I'll see what I can easily do. That logic is going to change soon as the information about level is going to come from browser-specs.

tidoust:
OK, I amended the CSS/IDL extract generation rule to be "Create the extract for the latest full level that defines relevant terms, or to the latest delta level if that's the only one that defined relevant terms", adjusted and improved the logic of the "isLatestLevel" function accordingly (now called "isLatestLevelThatPasses" because it takes a predicate that specs must pass to be considered), and added unit tests.

I'll move ahead and merge. Some CSS extracts will change as a result because there will be for the non-delta version of the spec. Existing IDL extracts should not change at all (and, thanks to the amendment, the crawl will continue to generate an IDL extract for CSS Images Level 4).

PR #235, opened 2020-01-27
Update URL of WebRTC Priority Control API

tidoust:
Shortname changed, see:
https://github.com/w3c/webrtc-priority/issues/8#issuecomment-578736858

Notes:
- The IDL extract will now be `webrtc-priority.idl` and not `webrtc-dscp.idl`.
- The previous shortname has become an alias from a W3C API perspective which made Reffy fail (the W3C API does not yet handle aliases).

I'll rename the IDL file in `reffy-reports`.


tidoust:
There should be no need to rename the IDL file in `reffy-reports` actually. That should happen automatically.

PR #248, opened 2020-02-22
Return normal strings from extraction functions

saschanaz:
No need to return Promises as nothing is really asynchronous here.


tidoust:
Absolutely, and I don't think that there ever was a reason to think that this could ever need to be asynchronous. Thanks for the simplification!

PR #254, opened 2020-02-29
Drop JSDOM and process specifications in Puppeteer

tidoust:
Specs crawled by Reffy were loaded using JSDOM. This has created a number of gotchas over time due to missing features in JSDOM that specifications (or specification authoring tools) actually need.

We notably had to switch to respecDocWriter recently to process Respec documents, which is based on Puppeteer under the hoods. But we were still using JSDOM to load initial specifications.

This update **completely removes the dependency on JSDOM**. Instead, the code now loads the specification in **Puppeteer**, and runs all extractions directly there.

To achieve this, the code had to be refactored:
- extraction logic now gets exposed as **ES6 modules**, defined in "src/browserlib"
- these modules get bundled with rollup.js and loaded in the browser context through Puppeteer (bundling is needed not to have to serve the script from a proper HTTPS server).
- a new "**processSpecification**" utility now replaces all loading methods. It allows to load a URL in Puppeteer and to run a callback function on the resulting document once it is done loading. The callback function has access to ES6 modules through the "**window.reffy**" namespace.
- CLI tools have been re-written accordingly.

Since Puppeteer runs a headless browser in a separate process, this completely removes the need to create child processes on our own, which further simplifies the crawling logic.

Loading the specification in Puppeteer means that we can also remove the dependency on respecDocWriter.

Note the need to bundle ES6 modules means that "npm run build" must now be run each time the source changes in "src/browserlib" to updates built files in the "builds" folder.

The code still behaves as before with regards to network requests, meaning that it intercepts network requests and serve them from the local cache storage when possible, or simply abort the requests when they don't seem needed (e.g. those on images). See comments at the top of the "processSpecification" method for details.

Once in a while - that seems hard to reproduce reliably - Puppeteer and/or our network handling logic leaves some resources open that prevent Node.js from exiting at the end of the crawl. A call to "process.exit(0)" is now made at the end of the processing to guarantee that the crawl will come to a halt.


tidoust:
@foolip, the pull request is a bit large, but most of it is really just code moving from one place to another. Main changes are in `src/lib/util.js` where the `processSpecification` method replaces the former spec loading logic, to use Puppeteer.

Running a crawl with this new version of the whole list of specs yields the same results (CSS/IDL extracts, but also results in crawl.json and anomalies reports) as with the previous one.

src/browserlib/get-title.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Gets the title of the document
+ */
+export default function () {
+  const title = window.document.querySelector('title');
+  if (window.location.href === 'https://html.spec.whatwg.org/multipage/workers.html') {
```
dontcallmedom: in the brave new MoU world, do we still need that exception? (probably no need to fix it here and now)

tidoust:
I'll go ahead and merge this to be able to work on further updates. Comments certainly still welcome if needed!

PR #282, opened 2020-04-10
Remove Box Tree API Level 1

foolip:
See discussion in https://github.com/web-platform-tests/wpt/pull/12805


foolip:
@tidoust this is really just to avoid attempting to export the IDL file to WPT. Is there value of scraping this spec outside of the IDL fragments, such that it should be kept?

Maybe we should have a "abandoned" or "no implementer interest yet" state to handle this?

Issue #303, opened 2020-05-15
Screen Wake Lock API URL is wrong


foolip:
https://www.w3.org/TR/wake-lock/ is the URL listed in specs-idl.json:
https://github.com/tidoust/reffy/blob/c579ea60d52e3a141f5e8d8d30b85a33a5e7c098/src/specs/specs-idl.json#L122

That copy links to https://w3c.github.io/wake-lock/, which redirects to https://w3c.github.io/screen-wake-lock/ using `<meta http-equiv="refresh" content="1;url=https://w3c.github.io/screen-wake-lock/" />`.

The resulting IDL file in reffy-reports seems to have the title of the new spec, but the old URL:
https://github.com/tidoust/reffy-reports/blob/fe5461acc8463ff16dfa9243e2f4e4d9e1ebde98/ed/idl/wake-lock.idl#L4

However, in https://github.com/tidoust/reffy-reports/blob/master/ed/crawl.json, it looks like the redirect wasn't successfully followed.

@tidoust how should we handle this? The easiest would be to just list https://w3c.github.io/screen-wake-lock/ instead of https://www.w3.org/TR/wake-lock/.

foolip:
I've sent https://github.com/tidoust/reffy/pull/304, but half expect that's not the best solution to this.

tidoust:
@deniak updated the URL of the Editor's Draft in W3C systems, which fixes the issue (picked up in https://github.com/tidoust/reffy-reports/commit/d030e72adf94b43af80b5595b51de5fb683159e5)

FWIW, Reffy normally manages to follow these redirects, but chokes on this one because of the 1 second pause. I crafted a PR to fix that for the sake of making it easier for machines to take over the world...

PR #304, opened 2020-05-15
Update URL for Screen Wake Lock API

foolip:
Fixes https://github.com/tidoust/reffy/issues/303.


tidoust:
Closing without merging as the update should no longer be needed, see discussion in #303.

PR #312, opened 2020-05-26
Switch to w3c/browser-specs for list of specs

tidoust:
The update makes Reffy use the list of specs exported by w3c/browser-specs as the initial list of specs to crawl. Lists of specs in `src/specs` have been dropped accordingly.

The update also means that Reffy no longer fetches spec info from the W3C API and from Specref, removing the need to store a W3C API key.

The find-spec CLI command was removed since it is no longer relevant. It should be moved to w3c/browser-specs.


tidoust:
@foolip This will enact the switch to [w3c/browser-specs](https:/github.com/w3c/browser-specs/). There are a few additional specs in browser-specs that were not yet in the list of specs crawled by Reffy. So merging this will trigger the creation of a few additional IDL extracts:

- [ScrollToTextFragment](https://wicg.github.io/ScrollToTextFragment/)
- [compression](https://wicg.github.io/compression/)
- [contact-api](https://wicg.github.io/contact-api/spec/)
- [css-highlight-api](https://drafts.csswg.org/css-highlight-api-1/)
- [get-installed-related-apps](https://wicg.github.io/get-installed-related-apps/spec/)
- [gpuweb](https://gpuweb.github.io/gpuweb/)
- [mathml-core](https://mathml-refresh.github.io/mathml-core/)
- [portals](https://wicg.github.io/portals/)
- [serial](https://wicg.github.io/serial/)
- [trusted-types](https://w3c.github.io/webappsec-trusted-types/dist/spec/)

All existing extracts should remain unchanged.

foolip:
That is excellent, more spec coverage is exactly what I was hoping for! Are any specs also lost in the process?

tidoust:
Switch should not trigger any loss, browser-specs contains all the specs that previously were in Reffy.

foolip:
Fantastic!

PR #319, opened 2020-06-03
Include the generator (Bikeshed, ReSpec, etc.) in crawl.json

foolip:
Just for fun, and very little wiring is needed to extract it.

### DIFF
--- a/src/cli/crawl-specs.js
+++ b/src/cli/crawl-specs.js
@@ -78,6 +78,7 @@ async function crawlSpec(spec, crawlOptions) {
             return {
                 crawled: window.location.toString(),
                 title: window.reffy.getTitle(),
+                generator: window.reffy.getGenerator(),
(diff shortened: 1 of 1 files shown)


foolip:
Running this locally, as expected there aren't any Anolis specs around any longer. If this is merged I can send a follow-up PR to remove the supporting code.

tidoust:
Forgot to add: indeed, Anolis detection code could be dropped!
Note that you'll need to re-generate `browser.js` through a call to `npm run build`.

foolip:
This took effect in https://github.com/tidoust/reffy-reports/commit/a171b482244476362031fd27f6f8024758c168ea#diff-564e73615fe0cfec1aeae49eb06211f0.

Looks like there are:
- 212 Bikeshed specs
- 66 ReSpec specs
- 18 specs with an unidentified generator

PR #320, opened 2020-06-03
Remove special support for Anolis specs

foolip:


src/browserlib/extract-references.js
```diff
@@ -51,12 +51,6 @@ function getExtractionRules(generator) {
         informative: "#informative-references > dl"
       }
     },
```
tidoust, line 53: Nit: Could you drop the trailing comma here?

foolip: Done (force pushed)

PR #337, opened 2020-06-24
Normalize whitespaces when extracting definitions

### LINKED ISSUE #335: dfn extraction: normalize whitespaces
See discussion in #332

tidoust:
Fixes #335


src/browserlib/extract-pages.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Extract pages from the table of contents of the current document.
+ *
```
dontcallmedom, line 3: sounds like out of scope of this pull request :)

tidoust: Indeed :) Dropped.

Issue #336, opened 2020-06-24
Refine Specifications Definitions data model


dontcallmedom:
Bikeshed defines the [contract that specifications need to follow](https://tabatkins.github.io/bikeshed/#dfn-contract) to have their definitions included in the bikeshed list of references, which [ReSpec imports](https://github.com/sidvishnoi/respec-xref-route/blob/master/scraper.ts#L16) as well.

Reffy aims to provide a [new source of data for these definitions](https://github.com/tidoust/reffy/pull/229) with more up-to-date and accurate data (given e.g. its capability to parse ReSpec live-generated editors draft), and while there are a few remaining issues (#333, #334, #335), the gist of the [diff between what Reffy finds and what Shepherd finds](https://github.com/tidoust/reffy/issues/332) comes from rules that Shepherd applies that aren't described in the contract.

This issue aims at gathering input from @tabatkins, @marcoscaceres @sidvishnoi and @plinss on refinements to the spec definition parsing contract, to make sure what reffy produces aligns with what editing tools outputs and what editing tools would want to consume for cross-references.

Analyzing the [diff between reffy and shepherd](https://github.com/tidoust/reffy/issues/332), the following aspects need clarification:
* in `dfn-type`  `dfn`, shepherd seems to convert curly quotes (`’`) into straight ones (`'`) - is that something we want to preserve?
* likewise, quotes that surround the whole linking text get removed by shepherd in many cases (e.g. enum values in general, HTML definitions e.g. ["non-blocking"](https://html.spec.whatwg.org/multipage/scripting.html#non-blocking)), but not all (e.g. ["strict-origin-when-cross-origin"](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-strict-origin-when-cross-origin)) - it would be best to clarify the intent in the parsing contract
  * with `dfn-type` `enum-value`, bikeshed includes a version of the linking text which includes the surrounding `"` , whereas respec only has the `"`-less version; shepherd strips them (per the above), but it would seem better to align the editing tools here
* right now, for functions (incl methods, constructors), bikeshed automatically exports definitions of the method with and without arguments; in respec, this is left to the discretion of spec editors, which create inconsistent results; in practice, I'm not sure the versions with arguments help anyone.
[...]

Issue #334, opened 2020-06-24
Adjust dfns extraction for SVG 2 spec


tidoust:
See discussion in #332.

Issue #333, opened 2020-06-24
Adjust dfns extraction for HTML spec


tidoust:
See discussion in #332. The HTML spec does not follow the usual contract.

dontcallmedom:
we now are beyond parity with shepherd, so closing - see also https://github.com/whatwg/html/pull/5694 for discussing on upstreaming the changes to the HTML spec

PR #339, opened 2020-06-25
Add tests for definition extraction

dontcallmedom:
to help with extending it to HTML (#333), SVG (#334) and refining the rules (#336)

### DIFF
--- a/tests/extract-dfns.js
+++ b/tests/extract-dfns.js
@@ -17,7 +17,7 @@ const tests = [
    html: "<dfn id='foo'>Foo</dfn>",
    dfns: [{}]
   },
-  {title: "normalize whitespace and trims text in a simple <dfn>",
+  {title: "normalizes whitespace and trims text in a simple <dfn>",
    html: "<dfn id='foo'>Foo\n    \n</dfn>",
    dfns: [{}]
   },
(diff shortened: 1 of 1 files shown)


tests/extract-dfns.js
```diff
@@ -0,0 +1,97 @@
...
+    for: [],
+    access: 'private',
+    informative: false
+};
+const tests = [
+  {title: "parses a simple <dfn>",
+   html: "<dfn id='foo'>Foo</dfn>",
+   dfns: [{}]
+  },
+  {title: "normalize whitespace and trims text in a simple <dfn>",
```
tidoust, line 20: s/normalize/normalizes

tests/extract-dfns.js
```diff
@@ -0,0 +1,97 @@
...
+    localLinkingText: [],
+    type: 'dfn',
+    for: [],
+    access: 'private',
+    informative: false
+};
+const tests = [
+  {title: "parses a simple <dfn>",
+   html: "<dfn id='foo'>Foo</dfn>",
+   dfns: [{}]
```
tidoust, line 18: I thought I'd note that I had some initial difficulty reading this because, without looking at the rest of the code, this suggests that the extraction will produce one empty object. That is, it would be easier to grasp immediately if here and below, this were `[baseDfn]` (or `[Object.assign({}, baseDfn)]`). I realize that means duplicating code, but then code is written once, read many times...

PR #340, opened 2020-06-29
Add parsing of definitions specific to HTML spec

dontcallmedom:
The HTML spec does not follow the markup conventions to properly type and associate definitions, so it needs special handling.
    
A lot is encoded in the ids of the definition elements, but not all the data (e.g. no distinction between attributes and methods), and there are many exceptions to the general conventions on id building.
    
Figuring out the right output requires additional knowledge that is available through the WebIDL fragments embedded in the spec, so we make use of that.

The diff of the generated definitions is large (2000+ definitions have a changed type), so a bit hard to meaningfully review. I think the result is statistically much better, so even if there are regressions hidden in the middle, that's probably still an overall improvement.

I'm separately looking into whether the said improved definitions can be upstreamed into the HTML spec.

### DIFF
--- a/src/browserlib/extract-dfns.js
+++ b/src/browserlib/extract-dfns.js
@@ -179,16 +179,16 @@ function preProcessHTML() {
   function fromIdToTypeAndFor(containerid, id) {
     // deals with exceptions to how containerid / id are expected to be parsed
     if (id) {
-    [containerid, id] = {
-      "history-scroll": ["history", "scrollrestoration"],
-      // overloads
-      "document-open" : ["document", "open"],
-      "dedicatedworkerglobalscope-postmessage": ["dedicatedworkerglobalscope", "postmessage"],
-      "messageport-postmessage": ["messageport", "postmessage"],
-      "window-postmessage": ["window", "postmessage"],
-      "worker-postmessage": ["worker", "postmessage"],
-      "context-2d-settransform": ["context-2d", "settransform"]
-    }[containerid] || [containerid, id];
+      [containerid, id] = {
+        "history-scroll": ["history", "scrollrestoration"],
+        // overloads
+        "document-open" : ["document", "open"],
+        "dedicatedworkerglobalscope-postmessage": ["dedicatedworkerglobalscope", "postmessage"],
+        "messageport-postmessage": ["messageport", "postmessage"],
+        "window-postmessage": ["window", "postmessage"],
+        "worker-postmessage": ["worker", "postmessage"],
+        "context-2d-settransform": ["context-2d", "settransform"]
+      }[containerid] || [containerid, id];
     }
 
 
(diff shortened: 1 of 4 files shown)


src/browserlib/extract-dfns.js
```diff
@@ -19,72 +19,451 @@
...
+    }
+
+    if (interfaces.length) {
+      let type = "attribute";
+      let relevantInterfaces = interfaces;
+      if (id) {
+        type = "dfn";
+        // dom-head-profile, intentionally omitted from IDL fragment
+        if (id === "profile" && containerid === "head") {
+          return {type: "attribute", _for:"HTMLHeadElement"};
```
tidoust, line 223: Nit: missing space after `_for:`, 8 occurrences in total, consistency FTW!

src/browserlib/extract-dfns.js
```diff
@@ -19,72 +19,451 @@
...
+        relevantInterfaces = interfaces.filter(iface => iface.members.find(member => member.name && member.name.toLowerCase() === id));
+        if (relevantInterfaces.length) {
+          let idlTerm = relevantInterfaces[0].members.find(member => member.name && member.name.toLowerCase() === id);
+          type = idlTerm.type === "operation" ? "method" : idlTerm.type;
+        }
+      }
+      return {type, _for: [... new Set(relevantInterfaces.map(iface => iface.name))].join(",")};
+    }
+
+    const enumName = id => { switch(id) {
```
tidoust, line 234: Could you move the `switch` to a new line, and move closing `}` back to "normal" place?

src/browserlib/extract-dfns.js
```diff
@@ -19,72 +19,451 @@
...
+        return;
+      }
+
+      if ((m = el.id.match(/^selector-/))) {
+        el.dataset.dfnType = 'selector';
+        return;
+      }
+
+      if ((m = el.id.match(/^dom-([^-]+)$/) || el.id.match(/^dom-([^-]+)-[0-9]+$/) || el.id.match(/^dom-([^-]+)-constructor$/))) {
+        const globalscopes = ["Window", "WindowOrWorkerGlobalScope", "HTMLElement", "WindowSessionStorage", "WorkerGlobalScope", "WindowLocalStorage", "HTMLOrSVGElement", "ElementContentEditable"]
```
tidoust, line 394: Nit: Sort A-Z with one per line

tests/extract-dfns.js
```diff
@@ -62,21 +96,253 @@ const tests = [
...
+  },
+  {title: "ignores elements that aren't <dfn> and headings",
+   html: "<span data-dfn-type=dfn id=foo>Foo</span>",
+   changesToBaseDfn: []
+  },
+  {title: "ignores headings without a data-dfn-type",
+   html: "<h2 id=foo>Foo</h2>",
+   changesToBaseDfn: []
+  },
+  {title: "includes data-lt in its list of linking text",
```
tidoust, line 131: Note it's not only "includes" as the attribute completely overrides the text content, replace with "uses text in data-lt as linking text"?

src/lib/util.js
```diff
@@ -297,6 +297,11 @@ async function processSpecification(spec, callback, args, counter) {
             path: path.resolve(__dirname, '../../builds/browser.js')
         });
 
+        // Import WebIDL2, needed to parse definitions out of the HTML spec
+        await page.addScriptTag({
```
tidoust, line 301: That should no longer be needed now that you use ES modules.

The WebIDL2.js library ends up in `browser.js`.

I note that this may not be such a fantastic idea after all. We'll have to remember to re-build `browser.js` whenever we bump the version of the WebIDL.js library. An alternative would be to define WebIDL2.js as a module here and tell Rollup to exclude it from the build. Anyway, I suggest to just proceed for now (dropping these lines).

But well, I suggest to just proceed for now.

PR #343, opened 2020-07-01
Fix handling of exceptional attribution of methods/attributes to interfaces

dontcallmedom:


tidoust:
The update makes the test "handles finding IDL type across mixins and partial" fail.

dontcallmedom:
yay for tests, boo for piles of fragile heuristics

PR #344, opened 2020-07-01
Handle SVG-specific conventions in definitions

### LINKED ISSUE #334: Adjust dfns extraction for SVG 2 spec

dontcallmedom:
This brings the output of Reffy to parity with shepherd on the SVG2 spec
close #334

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,6 @@
     "ed-html": "node reffy.js run ed html",
     "ed-diff": "node reffy.js run ed diff",
     "ed-diffnew": "node reffy.js run ed diffnew",
-    "test": "jasmine && mocha --recursive tests/"
+    "test": "npm run build && jasmine && mocha --recursive tests/"
   }
 }
--- a/package.json
+++ b/package.json
@@ -66,6 +66,6 @@
     "ed-html": "node reffy.js run ed html",
     "ed-diff": "node reffy.js run ed diff",
     "ed-diffnew": "node reffy.js run ed diffnew",
-    "test": "npm run build && jasmine && mocha --recursive tests/"
+    "test": "jasmine && mocha --recursive tests/"
   }
 }
--- a/src/browserlib/extract-dfns.js
+++ b/src/browserlib/extract-dfns.js
@@ -95,9 +95,13 @@ export default function (spec) {
   ].join(',');
 
   let extraDefinitions = [];
-
-  if (spec === "html") {
+  switch (spec) {
+  case "html":
     preProcessHTML();
+    break;
+  case "SVG2":
+    preProcessSVG2();
+    break;
   }
 
   return [...document.querySelectorAll(definitionsSelector)]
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -66,6 +66,6 @@
     "ed-html": "node reffy.js run ed html",
     "ed-diff": "node reffy.js run ed diff",
     "ed-diffnew": "node reffy.js run ed diffnew",
-    "test": "jasmine && mocha --recursive tests/"
+    "test": "npm run build && jasmine && mocha --recursive tests/"
```
tidoust, line 69: It's convenient to add it here, but I believe we still need a "test" command that does not build `browser.js`, so as to detect cases where we fail to include it in the pull request (for instance, with this change and assuming tests passed, the fact that `browser.js` wasn't in the initial PR #343 may have gone unnoticed).

Or we need to switch to a mode where `browser.js` is not tracked in the repo at all.

src/browserlib/extract-dfns.js
```diff
@@ -480,3 +484,81 @@ function preProcessHTML() {
...
+  }
+
+  [...document.querySelectorAll(".attrdef dfn[id]:not([data-dfn-type]):not([data-skip])")]
+    .forEach(el => {
+      el.dataset.dfnType = "element-attr";
+      const attrDesc = document.querySelector('[data-reffy-page$="attindex.html"] th span.attr-name a[href$="#' + el.id + '"]');
+      if (attrDesc) {
+          el.dataset.dfnFor = attrDesc.closest('tr').querySelector('td').textContent;
+      } else {
+        console.error("Could not find description for " + el.textContent);
```
tidoust, line 507: Shouldn't this (and other call below) throw instead? `console.error` messages are likely going to be ignored, e.g. drowned in other console messages produced by ReSpec.

tests/extract-dfns.js
```diff
@@ -321,13 +359,159 @@ const tests = [
...
+            "unknown",
+            "use",
+            "video"
+           ],
+      "access": "public"
+    }],
+    spec: "SVG2"
+  },
+  {
+    title: "identifes the link element definition in SVG2 spec",
```
tidoust, line 410: s/identifes/identifies/g (there are other occurrences)

tests/extract-dfns.js
```diff
@@ -321,13 +359,159 @@ const tests = [
...
+      id: "__svg__SVGNameList__initialize",
+      linkingText: ["initialize"],
+      type: "method",
+      "for": ["SVGNameList"],
+      access: "public"
+    }],
+    spec: "SVG2"
+  },
+  {
+    title: "identified IDL interfaces in headings of the SVG2 spec",
```
tidoust, line 469: s/identified/identifies

tests/extract-dfns.js
```diff
@@ -321,13 +359,159 @@ const tests = [
...
+    spec: "SVG2"
+  },
+  {
+    title: "identifes attributes defined with class adef in SVG2 spec",
+    html: `<dt id="PatternElementPatternUnitsAttribute"><span class="adef">patternUnits</span></dt>`,
+    changesToBaseDfn: [{
+      id: "PatternElementPatternUnitsAttribute",
+      linkingText: ["patternUnits"],
+      type: "element-attr",
+      "for": ["pattern"],
```
tidoust, line 427: Why `"for"` and not `for` as elsewhere?
(3 occurrences)

package.json
dontcallmedom: I guess it is my overall preference not to include built files in repos, but I know this isn't a shared preference, and there may be good reasons to have it included here.

In any case, you're right that this is out of scope for this PR, so I will remove it

tidoust: I don't disagree. I think I initially included the build to be able to keep `rollup` as a `devDependencies`, and to be able to run Reffy commands right away without having to build anything first. But it's not a big deal to move `rollup` to the `dependencies` list (that won't be the largest one ;)), and we can easily detect the absence of `browser.js` in the code to report a nice error message, or even to run the build from the code.

I just want to avoid is a situation where parts of the code assume that the build is there while other parts don't :)

src/browserlib/extract-dfns.js
dontcallmedom: if we throw and the SVG spec is updated "incorrectly", what will happen? will reffy updates be blocked until this gets fixed?

(I note that in HTML spec parsing I used yet another approach of setting type "unknown"; this should be aligned in any case)

PR #350, opened 2020-07-03
Add extraction of headings to reffy

dontcallmedom:
part of #346

### DIFF
--- a/src/browserlib/extract-headings.js
+++ b/src/browserlib/extract-headings.js
@@ -1,0 +1,9 @@
+/**
+ * Extract headings data from documents
+*/
+export default function () {
+  const headings = {};
+  return [...document.querySelectorAll('h1[id], h2[id], h3[id], h4[id], h5[id] ,h6[id]')].map(n => {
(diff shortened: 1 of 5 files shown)


src/browserlib/extract-headings.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Extract headings data from documents
+*/
+export default function () {
+  const headings = {};
+  return [...document.querySelectorAll('h1[id], h2[id], h3[id], h4[id], h5[id] ,h6[id]')].map(n => {
+    return {id: n.id, level: n.tagName.slice(1), title: n.textContent.trim().replace(/^[0-9\.]+ /, '').trim()};
```
tidoust, line 7: Have you checked whether levels are always meaningful in our specs?

In theory, there could be nested `<section>` that simply reuse `<h1>` throughout. In practice, I suspect editing tools use appropriate headings.

dontcallmedom: I haven't - I'm assuming it will be easier to do once we collect the data :)

dontcallmedom:
>  Not enabled for now on purpose, I suppose?

no, that was an oversight :) fixed, but that means a (simple) conflict to resolve

PR #351, opened 2020-07-03
Pass argument to extractDefinitions (for good this time)

dontcallmedom:
Also, export headings in crawl output


dontcallmedom:
merging as hot-fix, since currently reffy-reports is broken, and next run is in 2 hours

PR #352, opened 2020-07-03
Add test to ensure the crawl runs as expected on mock content

dontcallmedom:


src/cli/crawl-specs.js
```diff
@@ -163,7 +163,7 @@ async function crawlSpec(spec, crawlOptions) {
  * @return {Promise<Array(Object)} The promise to get an array of complete
  *   specification descriptions
  */
-async function crawlList(speclist, crawlOptions, resultsPath) {
+async function crawlList(speclist, crawlOptions) {
```
tidoust, line 166: Indeed, the parameter is useless. Since you're at it, could you drop the parameter from the function call on line 395?

src/lib/util.js
```diff
@@ -235,7 +235,7 @@ async function processSpecification(spec, callback, args, counter) {
             for (const url of pageUrls) {
                 const subAbort = new AbortController();
                 const subPage = await browser.newPage();
-                const subCdp = await page.target().createCDPSession();
+                const subCdp = await subPage.target().createCDPSession();
```
tidoust, line 238: Good catch :)

PR #353, opened 2020-07-03
Use browser-spec data to parse subpages

### LINKED ISSUE #342: Drop logic to identify pages in multi-page specs
Information about pages that compose a multi-page spec is now available from browser-specs, no need to implement that logic in Reffy anymore.

dontcallmedom:
close #342
builds on #352


dontcallmedom:
I made the PR on the `test-crawl` branch, but the goal will be to get it merged in `master`

src/cli/crawl-specs.js
```diff
@@ -63,8 +63,8 @@ const byURL = (a, b) => a.url.localeCompare(b.url);
  */
 async function crawlSpec(spec, crawlOptions) {
     spec.crawled = crawlOptions.publishedVersion ?
-        (spec.release ? spec.release.url : spec.nightly.url) :
-        spec.nightly.url;
+        (spec.release ? spec.release : spec.nightly) :
+        spec.nightly;
```
tidoust, line 67: I must be missing something. Wouldn't that change `crawled` from being a URL to being an object with a `url` property (and possibly other properties such as `pages`)? Is that intended?

dontcallmedom: yes, that's intended - `processSpecification` accepts either a URL or an object, and I need to pass the information about pages to replace the current in-line detection of pages.

tidoust: Then that's going to affect report generation, which expects `crawled` to be a URL:
https://github.com/tidoust/reffy/blob/master/src/cli/generate-report.js#L73

I think that's the only place that needs fixing.

dontcallmedom: the `crawled` that appears in the final data is set as a string on L79, so it shouldn't be affected by the value of the method parameter (which in any case is normalized to an object in the very first line of the `processSpecification` function.

This is also confirmed by the fact that `tests/crawl-test.json` doesn't show a modified `crawled` field in its output.

tidoust: I see. It's not fantastic to use the "same" property to store different types of values but I'll rewrite later on. Property assignments are a bit messy in this function, precisely because it's not easy to tell what comes from elsewhere, and what gets overwritten.

dontcallmedom:
I didn't know, but it was in fact [quite easy](https://docs.github.com/en/enterprise/2.14/user/articles/changing-the-base-branch-of-a-pull-request)

Issue #386, opened 2020-08-27
check-specs utility needs fixing


tidoust:
The check-specs utility, that takes a spec or a list of specs as input, crawls them, and analyzes the result against a reference crawl report, is broken, likely due to the switch to browser-specs.

Issue #385, opened 2020-08-26
Get rid of crawl.json


tidoust:
Crawl results end up in `crawl.json`. The file is already >26MB (and yet it does not even contain raw IDL extracts). We're adding more properties to the crawl over time. For instance, we recently added definitions and targeted fragments in external documents. The list of properties should continue to grow over time, e.g. see #153.

Having one big `crawl.json` file is not fantastic. It makes it hard for humans to look for something in the file, which is often useful, especially when debugging. This includes making it hard to review diffs and spot crawl errors. Also, editing tools tend to work better with smaller JSON files. For instance, GitHub won't show the contents of the file because it is too large.

We're already producing separate CSS definitions, definitions, and IDL extract files for each spec. The idea would be to replace `crawl.json` with an index file that only contains generic information about crawled specs (metadata from w3c/browser-specs, possible error message if crawl went wrong, and links to individual files that got created during the crawl). The index file would look like:

```json
{
  "type": "crawl",
  "title": "Reffy crawl",
  "date": "2020-08-26T06:37:45.423Z",
  "options": {},
  "stats": {
    "crawled": 331,
    "errors": 0
  },
  "results": [
    {},
    {}
  ]
```

Individual results would look like:

```json
{
  "url": "https://compat.spec.whatwg.org/",
  "crawled": "https://compat.spec.whatwg.org/",
  "status": "OK",
  "seriesComposition": "full",
  "shortname": "compat",
  "series": {
    "shortname": "compat",
    "currentSpecification": "compat"
  },
  "nightly": {
    "url": "https://compat.spec.whatwg.org/",
    "repository": "https://github.com/whatwg/compat",
    "filename": "index.html"
  },
  "title": "Compatibility Standard",
  "source": "specref",
  "shortTitle": "Compatibility",
  "date": "29 June 2020",
  "generator": "bikeshed",
  "links": "links/compat.json",
  "refs": "refs/compat.json",
  "idl": "idl/compat.idl",
  "idlparsed": "idlparsed/compat.json",
  "css": "css/compat.json",
  "dfns": "dfns/compat.json",
  "headings": "headings/compat.json"
}
```

More properties may be added over time.

External projects may depend on `crawl.json`, so we may want to keep it for a while, and create a separate `index.json` file instead.

Note that, at least for now, CSS definitions and raw IDL extracts are only produced for the latest version of a spec in a series (that eases ingestion in Web Platform Tests, which follows the same model). Other files are generated per spec, not per series.

TODO:
[...]

PR #388, opened 2020-08-27
Generate index file and additional individual extracts

tidoust:
This update partially addresses #385. It updates the code to generate an `index.json` file as described in the issue, where spec entries only contain metadata and links to generated dumps.

Additional individual JSON extracts are also generated from the info that was in `crawl.json` in subfolders that did not exist before:
- `idlparsed` for the parsed IDL
- `links` for the list of links (and fragments) to external documents
- `headings` for the list of headings
-  `refs` for the list of normative and informative references

These JSON files are per spec, named after the spec's shortname, and start with a short `spec` property that contains the title of the spec and the crawled URL to make it easier to consume the files without having to go through `index.json`.

The study tool no longer uses `crawl.json` but rather uses the `index.json` file directly. The anomalies report generator tool was restricted to only accept study files as input, and not also `crawl.json` to generate diffs: the ability to use `crawl.json` should not really be needed, starting from `study.json` makes things easier and also avoids having to fetch 27MB of data that would be only useful to re-create the `study.json` file.

The `crawl.json` file is still generated as before. To be dropped once we're confident that external projects no longer depend on it.

Note: the `check-specs` CLI is actually broken, see #386. File update in this commit does not fix that.


src/cli/crawl-specs.js
```diff
@@ -239,9 +239,47 @@ async function saveResults(crawlOptions, data, folder) {
...
+                const json = JSON.stringify(contents, null, 2);
+                const filename = path.join(folders[property], spec.shortname + '.json');
+                try {
+                    await fs.promises.writeFile(filename, json);
+                }
+                catch (err) {
+                    // TODO: report error!
+                    console.log(err);
+                }
+                spec[property] = `${property}/${spec.shortname}.json`;
```
dontcallmedom, line 276:
```suggestion
                spec[property] = `${folders[property]}/${spec.shortname}.json`;
```

dontcallmedom (approved):

PR #393, opened 2020-09-02
Update URL in IDL preamble following repo transfer

tidoust:
URL used to point at tidoust/reffy-reports. The update makes it point at the new w3c/webref repository.

Update will change all generated IDL files at once. Merge should to be done in sync with @foolip to avoid polluting Web Platform Tests with hundreds of editorial PRs...


foolip:
I've sent https://github.com/web-platform-tests/wpt/pull/25367 to match this. It looks like I have write access to this repo, so I can merge both when approved.

foolip:
I've merged the two PRs just now.

PR #407, opened 2020-09-29
Create CSS/IDL files for delta specs as well

tidoust:
As opposed to other extracts that are generated per spec, CSS and IDL extracts are generated per series. This update adds CSS and IDL extracts for delta specs. See related request in https://github.com/w3c/webref/issues/60

The extracts for the delta spec are named after the specification's shortname, and not after the specification's series' shortname. In other words, the update will create `idl/css-animations-2.idl`, not `idl/css-animations-delta.idl`.

@foolip, the reasoning behind the naming is that there may come a time when WPT needs to track more than 2 versions of a given spec at once. Using the spec's shortname, that's easy to do. If we rather use a `-delta` suffix, we'd have to rename things. Now, this may never happen, so let me know if you would rather prefer to see names that end with `-delta`.

### DIFF
--- a/src/cli/crawl-specs.js
+++ b/src/cli/crawl-specs.js
@@ -281,7 +281,7 @@ async function saveResults(crawlOptions, data, folder) {
         };
     }
 
-    async function saveIdl(spec) {
+    async function saveIdl(spec, filename) {
         let idlHeader = `
             // GENERATED CONTENT - DO NOT EDIT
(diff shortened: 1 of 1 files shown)


dontcallmedom (approved):

foolip:
As long as there are no cases where the part before `-2` or `-delta` differ, then it doesn't really matter I think, it's just a filename.

Thanks for fixing this!

PR #418, opened 2020-10-09
Ignore all files/dirs within reports/*/ except for README.md

gsnedders:
The list of files in `.gitignore` currently doesn't actually cover everything we have nowadays. Regardless, it's unlikely that we're going to want to add further files here within this repository, so simply ignore everything without subdirectories of `reports` except for the `README.md` files.


tidoust (approved):
Nice! I didn't even know that the `!` operator existed for gitignore...

PR #419, opened 2020-10-09
Find CSS properties that aren't in the tables

gsnedders:
I'm not really sure if this is a good idea (versus just having people look through `dfns` to get the complete list of all properties), but I wrote this so it exists now 🤷🏻 Opinions welcome, etc.


tidoust:
The CSS extracts should contain all CSS properties and, ideally, there should be no diff between the list of CSS properties in the CSS extract and the list of CSS properties in the dfn extract. In other words, this pull request is needed :)

This code would add the following properties:

1. [`stop-color`](https://svgwg.org/svg2-draft/pservers.html#StopColorProperty) and [`stop-opacity`](https://svgwg.org/svg2-draft/pservers.html#StopOpacityProperty) in SVG. It would actually be good to add custom extraction code for these definitions. That can be done separately though.
2. All [`-webkit-*` property aliases](https://compat.spec.whatwg.org/#css-simple-aliases) in Compat.
3. [Legacy Gap Properties](https://drafts.csswg.org/css-align-3/#gap-legacy) in CSS Box Alignment.
4. The legacy [`-webkit-user-select`](https://drafts.csswg.org/css-ui-4/#propdef--webkit-user-select) and [`-webkit-appearance`](https://drafts.csswg.org/css-ui-4/#propdef--webkit-appearance) properties in CSS Basic User Interface.
5. The legacy [`-webkit-line-clamp`](https://drafts.csswg.org/css-overflow-3/#propdef--webkit-line-clamp) property in CSS Overflow.

That all looks good.

One problem is that this would also add the [`property-name`](https://drafts.csswg.org/css2/#propdef-property-name) example in CSS 2, and the [`--*`](https://drafts.csswg.org/css-variables/#propdef-) family of properties in CSS Variables. These properties don't really exist, so it does not make sense to include them in the CSS extracts. They are [hardcoded in the CSS extraction code](https://github.com/tidoust/reffy/blob/master/src/browserlib/extract-cssdfn.js#L101). They probably have to be hardcoded here as well. Could you update the pull request accordingly? (The specs themselves should perhaps be updated since the definitions should actually not really be flagged as "property" `dfns`).

Out of curiosity, I also checked whether all CSS properties in the CSS extracts have corresponding entries in the dfns extract. That seems to be the case!

gsnedders:
> stop-color and stop-opacity in SVG. It would actually be good to add custom extraction code for these definitions. That can be done separately though.

Or someone should update the spec! 🙃

> (The specs themselves should perhaps be updated since the definitions should actually not really be flagged as "property" dfns).

Alas, given they're generated the same way as the other propdef my brief look at avoiding this in Bikeshed was… difficult. But probably worth filing an issue.

tidoust:
> One problem is that this would also add the property-name example in CSS 2, and the --* family of properties in CSS Variables. These properties don't really exist, so it does not make sense to include them in the CSS extracts. They are hardcoded in the CSS extraction code. They probably have to be hardcoded here as well. Could you update the pull request accordingly?

I updated the pull request to exclude `property-name` and `--*` from the CSS extracts.

Issue #413, opened 2020-10-08
Add closest heading under which a dfn is found


tidoust:
Reffy extracts definitions and headings. However, there is no way to tell where a particular definition was found in the spec. It could be useful to document the closest heading's ID.

For a possible usage, see https://github.com/mdn/browser-compat-data/issues/6765

PR #435, opened 2020-11-02
Add a heading property to dfns that target enclosing section

tidoust:
This update adds a `heading` property to dfns extracts as discussed in #413. This new property allows to answer the question: "Where does a definition appear in the spec?".

The `heading` property is an object with 3 keys:
- `id`: either the ID of the section, if there exists an explicit section element and if that section has an ID, or the ID of the heading element. The ID may be null is neither the section nor the heading element has an ID, although that should never happen with "modern" specs since authoring tools automatically generate these IDs.
- `title`: the title of the closest section's heading element, without the heading number. E.g. "Interface PresentationReceiver"
- `number`: The heading number, when heading is numbered, e.g. "6.6". Final "." is omitted (e.g. "A" and not "A.").

Under the hoods, this update adds the `createOutline` and `mapIdsToHeadings` functions to Reffy's crawling namespace. The first one computes the proper outline of the specification, following the "creating an outline" algorithm defined in HTML. The second one builds on top of the first one to map elements that have an ID to the logical section that contains them.

This update also modifies the headings extracts to reuse the result of the `mapIdsToHeadings` function. The net benefit is that the ID of the heading is now the heading of the section when it's defined, as done in the Table of Contents. For instance, the ID of the heading "A. IDL Index" in a ReSpec spec becomes "idl-index" and not "a-idl-index" (which is the ID of the heading element itself).

(As a side effect, heading's appendix titles are now also correctly parsed).

This update also removes duplicate dfns extraction tests.

One possible question: should the `heading` property merely point at the ID of the section? The ID can be associated with a title and number by looking at the headings extract. That said, having the info in the dfns extract directly, as done in this PR, looks handy.


src/browserlib/reffy.js
```diff
@@ -25,6 +27,8 @@ window.reffy = Object.assign(
     extractReferences,
     extractLinks,
     canonicalizeUrl,
-    canonicalizesTo
+    canonicalizesTo,
+    createOutline,
```
dontcallmedom, line 31: do we need to expose `createOutline` in the reffy object? as far as I can see, this is only used in the context of the test, which I assume can be avoided.

tests/extract-dfns.js
```diff
@@ -137,38 +137,6 @@ const tests = [
...
-   changesToBaseDfn: [{}]
-  },
-  {title: "ignores elements that aren't <dfn> and headings",
-   html: "<span data-dfn-type=dfn id=foo>Foo</span>",
-   changesToBaseDfn: []
-  },
-  {title: "ignores headings without a data-dfn-type",
-   html: "<h2 id=foo>Foo</h2>",
-   changesToBaseDfn: []
-  },
```
dontcallmedom, line 171: why were these removed?

src/browserlib/create-outline.js
```diff
@@ -0,0 +1,352 @@
+/**
+ * Creates an outline for the DOM subtree rooted at the given sectioning content
+ * or sectioning root element.
+ *
```
dontcallmedom, line 4: can we re-use an existing module instead? e.g. https://h5o.github.io/
if not, can we move this into a separate module? this doesn't feel reffy-esque since it's not spec-specific as far as I can tell.

src/browserlib/reffy.js
tidoust: It can be avoided. It is a bit of a pain, because tests run in `about:blank` pages, and there's no way to import a JS module directly in such pages (the browser cannot resolve `import` paths), so one has to convert the JS module to a regular JS function first. Anyway, I now use Rollup to do that conversion at the beginning of the test.

tests/extract-dfns.js
tidoust: "This update also removes duplicate dfns extraction tests"
I only removed them once, they're still in there!

src/browserlib/create-outline.js
tidoust: First, I wish I would have found that module earlier on, that might have saved me some coding time :) I see that I ended up with very similar code as in that module, which seems good.

Now, I don't see any easy way to re-use it because the creation of the outline is not what we're interested in. We're interested in mapping elements with IDs to their enclosing section. This mapping gets done during the generation of the outline, and this is what the createOutline function returns (along with the outline). The h5o module only creates the outline and wouldn't give us the information for non heading elements.

There may be a more direct way to get that information without creating an outline. Starting from an element with an ID, it seems doable to parse the DOM tree before it (in document order) to find the closest heading for it, taking into account sectioning rules in HTML. Given the way these sectioning rules are written in the HTML spec, it just seemed easier to follow the instructions and create the outline...

The code is indeed generic and could be moved to a separate library. But that means more work, so I wouldn't mind keeping it in Reffy for now.

dontcallmedom (approved):
great piece of work, thanks!

PR #444, opened 2020-11-13
Rename void to undefined in WebIDL analyser

dontcallmedom:


tidoust (approved):
We don't have many tests for the WebIDL parser, but that seems like a good occasion to add some for well-known types, c.f. `well-known.js` test file I added to the PR.

That looks good otherwise. Tests fail for now because respec.org is down. I believe that can be ignored.

PR #451, opened 2020-11-26
Add support for extracting ids from specs

dontcallmedom:
This will help detect links from one spec to another that don't exist - part of #346


tidoust (approved):
That seems good. Instinctively, I would have loved to capture some text too, but that seems hard to do in the generic case (for instance, the `id` may be on a `<section>` and collecting `textContent` would just create noise).

PR #455, opened 2020-12-04
Add script to generate report on spec links robustness

### LINKED ISSUE #346: Analyse backlinks to definitions
We already extract links that appear in specs (although we don't keep fragments for now). Goal is to keep fragments and compare with definitions to analyse:
1. cross-refs to definitions that are or are not flagged with a `data-export`
2. the list of specs that depend on a particular definition and that would "break" should the definition change.

dontcallmedom:
close #346 
The report generates list of links that either:
* Link to anchors that don't exist
* Link to anchors that no longer exist in the editor draft of the target spec
* Link to anchors that are not definitions or headings
* Link to definitions that are not exported
* Link to dated TR URLs
* Link to specs that should no longer be referenced

It also flags links that look like specs but that aren't recognized in our crawl (although most of these are false positives afaict).

### DIFF
--- a/src/cli/study-backrefs.js
+++ b/src/cli/study-backrefs.js
@@ -32,6 +32,7 @@ const shortNamesOfOutdatedSpecs = {
 };
 
 const shortnameMap = {
+  "mixedcontent": "mixed-content",
   "powerfulfeatures": "secure-contexts",
   "hr-time-2": "hr-time",
   "csp": "CSP",
@@ -145,11 +146,9 @@ function studyCrawlResults(results) {
               // Ignore links to previous versions
               if (m[1] !== spec.shortname) {
                 recordAnomaly(spec, "datedUrl", l);
-                return;
               }
             } else {
               recordAnomaly(spec, "unknownSpec", l);
-              return;
             }
           }
         }
@@ -158,16 +157,15 @@ function studyCrawlResults(results) {
         } else if (shortNamesOfOutdatedSpecs[shortname]) {
           recordAnomaly(spec, "outdatedSpec", l);
         }
+        if (!shortname) { return ;}
         // self-references might be broken because of ed vs tr
         if (shortname === spec.shortname || shortname === spec.series.shortname) return [];
         let sourceSpec = results.find(s => s.shortname === shortname || s.series.shortname === shortname);
         if (!sourceSpec) {
-          if (!shortNamesOfOutdatedSpecs[shortname] && !shortnameOfNonNormativeDocs.includes(shortname)) {
+          if (shortname && !shortNamesOfOutdatedSpecs[shortname] && !shortnameOfNonNormativeDocs.includes(shortname)) {
             recordUnknownSpec(l, shortname, spec)
           }
           return;
-        } else {
-          
         }
         let headings = sourceSpec.headings || [];
         let dfns = sourceSpec.dfns || [];
--- a/src/cli/study-backrefs.js
+++ b/src/cli/study-backrefs.js
@@ -1,15 +1,90 @@
 #!/usr/bin/env node
 const requireFromWorkingDirectory = require('../lib/util').requireFromWorkingDirectory;
-const canonicalizeUrl = require('../../builds/canonicalize-url').canonicalizeUrl;
-const canonicalizesTo = require('../../builds/canonicalize-url').canonicalizesTo;
-// FIXME
-const computeShortName = require('../../../browser-specs/src/compute-shortname');
+const {canonicalizeUrl, canonicalizesTo} = require('../../builds/canonicalize-url');
 const fs = require("fs");
 
 const matchSpecUrl = url => url.match(/spec.whatwg.org/) || url.match(/www.w3.org\/TR\/[a-z0-9]/) || (url.match(/.github.io/) && ! url.match(/w3c.github.io\/test-results\//));
 
-// TODO
-// report broken links based on crawled ids
+/*
+ TODO: DRY
+ Copied from browser-specs/src/compute-shortname.js
+*/
+function computeShortname(url) {
+  function parseUrl(url) {
+    // Handle /TR/ URLs
+    const w3cTr = url.match(/^https?:\/\/(?:www\.)?w3\.org\/TR\/([^\/]+)\/$/);
+    if (w3cTr) {
+      return w3cTr[1];
+    }
+
+    // Handle WHATWG specs
+    const whatwg = url.match(/\/\/(.+)\.spec\.whatwg\.org\/?/);
+    if (whatwg) {
+        return whatwg[1];
+    }
+
+    // Handle TC39 Proposals
+    const tc39 = url.match(/\/\/tc39\.es\/proposal-([^\/]+)\/$/);
+    if (tc39) {
+        return "tc39-" + tc39[1];
+    }
+
+
+    // Handle Khronos extensions
+    const khronos = url.match(/https:\/\/www\.khronos\.org\/registry\/webgl\/extensions\/([^\/]+)\/$/);
+    if (khronos) {
+        return khronos[1];
+    }
+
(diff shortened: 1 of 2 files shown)


tidoust (commented):
I'll re-write the script slightly. I understand that it is convenient to take `crawl.json` as input but the goal is to drop the file from the repo for good, so the script should rather start from `index.json`.

The function [`expandCrawlResults`](https://github.com/w3c/reffy/blob/master/src/lib/util.js#L395) is precisely meant to re-create `crawl.json` in memory from an `index.json` file.

src/cli/study-backrefs.js
```diff
@@ -0,0 +1,358 @@
...
+  missingSpecs[shortname].push({link, spec: spec.url});
+}
+
+function studyCrawlResults(edResults, trResults) {
+  edResults.forEach(spec => {
+    Object.keys(spec.links)
+      .filter(matchSpecUrl)
+      .forEach(l => {
+        let shortname;
+        let nakedLink = l.replace(/#.*$/, '');
```
tidoust, line 209: By definition, we store fragments separately, so that replacement is useless.

src/cli/study-backrefs.js
```diff
@@ -0,0 +1,358 @@
...
+              // Ignore links to previous versions
+              if (m[1] !== spec.shortname) {
+                recordAnomaly(spec, "datedUrl", l);
+              }
+            } else {
+              recordAnomaly(spec, "unknownSpec", l);
+            }
+          }
+        }
+        if (shortnameMap[shortname]) {
```
tidoust, line 232: These checks should only be meaningful when `computeShortname` is called. I suggest to move them there for clarity.

By definition, the contents of the crawl report should only contain "good" shortnames since we're going to use the crawl list as source to look for the spec.

src/cli/study-backrefs.js
```diff
@@ -0,0 +1,358 @@
...
+  "resource-timing-1": "resource-timing",
+  "wai-aria-1.1": "wai-aria-1.2"
+};
+
+// TODO: check the link is non-normative (somehow)
+const shortnameOfNonNormativeDocs = [
+  "aria-practices", "rdf11-primer", "discovery-api", "wake-lock-use-cases", "capability-urls", "streamproc", "media-source-testcoverage", "using-aria", "wai-aria-practices-1.1", "sensor-polyfills", "sensors", "design-principles", "security-questionnaire", "css3-preslev", "clreq", "klreq", "typography", "ssml-sayas", "css3-marquee", "spatial-navigation", "ilreq", "css-print", "books", "dpub-pagination", "predefined-counter-styles", "css-2017", "sniffly", "wai-aria-practices-1.2", "wai-aria-implementation", "wai-aria-roadmap", "wai-aria-practices", "webdatabase", "installable-webapps", "motion-sensors", "file-system-api", "media-accessibility-reqs", "webrtc-interop-reports", "webaudio-usecases", "web-audio-perf", "Audio-EQ-Cookbook", "web-intents", "touch-events-extensions", "fingerprinting-guidance", "webrtc-nv-use-cases", "html-design-principles", "storage-partitioning", "jlreq", "accept-encoding-range-test", "security-privacy-questionnaire", "dpub-latinreq"
+];
+
+const report = {};
```
tidoust, line 176: It would be better not to create global variables and rather move that within `studyCrawlResults`, in case we make this a real module later on.

src/cli/study-backrefs.js
```diff
@@ -0,0 +1,358 @@
...
+        } else if (shortNamesOfOutdatedSpecs[shortname]) {
+          recordAnomaly(spec, "outdatedSpec", l);
+        }
+        if (!shortname) { return ;}
+        // self-references might be broken because of ed vs tr
+        if (shortname === spec.shortname || shortname === spec.series.shortname) return [];
+        let sourceSpec = edResults.find(s => s.shortname === shortname || s.series.shortname === shortname);
+        let trSourceSpec = trResults.find(s => s.shortname === shortname || s.series.shortname === shortname);
+        if (!sourceSpec) {
+          if (shortname && !shortNamesOfOutdatedSpecs[shortname] && !shortnameOfNonNormativeDocs.includes(shortname)) {
```
tidoust, line 243: `shortname` is always set at this step, no need to check on it.

src/cli/study-backrefs.js
```diff
@@ -0,0 +1,358 @@
+#!/usr/bin/env node
+const requireFromWorkingDirectory = require('../lib/util').requireFromWorkingDirectory;
+const {canonicalizeUrl, canonicalizesTo} = require('../../builds/canonicalize-url');
+const fs = require("fs");
+
+const matchSpecUrl = url => url.match(/spec.whatwg.org/) || url.match(/www.w3.org\/TR\/[a-z0-9]/) || (url.match(/.github.io/) && ! url.match(/w3c.github.io\/test-results\//));
```
tidoust, line 6: What about `.csswg.org`, `.fxtf.org` and `.css-houdini.org` URLs?
Also, did you exclude Khronos Group specs on purpose?

src/cli/study-backrefs.js
```diff
@@ -0,0 +1,358 @@
...
+          return;
+        }
+        let headings = sourceSpec.headings || [];
+        let dfns = sourceSpec.dfns || [];
+        let ids = sourceSpec.ids || [];
+
+        // anchors
+        const anchors = spec.links[l];
+        for(let anchor of anchors) {
+          let id = ids.includes(anchor);
```
tidoust, line 255: Calling a boolean `id` is a good way to obfuscate the code...

src/cli/study-backrefs.js
```diff
@@ -0,0 +1,358 @@
+#!/usr/bin/env node
+const requireFromWorkingDirectory = require('../lib/util').requireFromWorkingDirectory;
+const {canonicalizeUrl, canonicalizesTo} = require('../../builds/canonicalize-url');
+const fs = require("fs");
```
tidoust, line 4: Not used. Can be dropped.

tidoust (commented):
See inline for a few additional comments. As said, I'm preparing an update, where I'll address them. One question remains on the `matchSpecUrl` function though.

I will also re-order some of the lines in a way that makes them more readable (at least for me :)).

src/cli/study-backrefs.js
dontcallmedom: I didn't (or at least I don't recall picking that subset on purpose) - I guess the list of bad links will grow quite a bit more…

src/cli/study-backrefs.js
dontcallmedom: always happy to help

src/cli/study-backrefs.js
```diff
@@ -0,0 +1,483 @@
...
+          nakedLink += '/';
+        }
+        shortname = (edResults.find(r =>
+          r.url === nakedLink ||
+          (r.release && r.release.url === nakedLink) ||
+          r.nightly.url === nakedLink ||
+          (r.series && nakedLink === 'https://www.w3.org/TR/' + r.series.shortname + '/') ) || {}).shortname;
+        if (!shortname) {
+          try {
+            ({shortname} = computeShortname(link));
```
tidoust, line 300: Talking about code obfuscation, this line does a very good job at pretending that it does something but actually doing something else. The `computeShortname` function returns a string, so this always sets `shortname` to undefined, which is probably not what was intended.

The logic of the few next lines remained somewhat good, except that the changes I made to make things more logic actually broke it. Anyway, I'll fix that.

tidoust:
I updated the script, expanding it to also check links to CSS drafts. I did not include links to Khronos group specs at this stage because some work seems needed to distinguish between links to specs and other types of resources under www.khronos.org.

dontcallmedom:
thanks a lot for your rewrite, great improvements :)

ready to merge from my perspective

PR #457, opened 2020-12-07
Limit files included in package

dontcallmedom:
could you also npm publish the new package and/or add me as maintainer of the package on npm?

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,12 @@
     "type": "git",
     "url": "https://github.com/w3c/reffy.git"
   },
+  "files": [
+    "index.js",
+    "reffy.js",
+    "src/",
+    "builds/"
+  ],
   "bugs": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -6,6 +6,10 @@
     "type": "git",
     "url": "https://github.com/w3c/reffy.git"
   },
+  "files": [
+    "index.js",
+    "src/"
```
tidoust, line 11: Won't `builds` be needed as well for `crawlSpecs`?

Also, since all `src/` script files will be in the pakage, why not include `reffy.js` as well?

tidoust:
I added you as maintainer of the package.

package.json
dontcallmedom: thanks, fixed

tidoust (approved):

PR #458, opened 2020-12-10
Drop builds folder, generate code dynamically

### LINKED ISSUE #387: Generate builds on the fly
The `builds` folder is in the repo, but shouldn't be as the files can be automatically generated, and it's easy to forget to update the builds when the source files change.

### LINKED ISSUE #415: Use a single browser instance
Running locally, it appears the runtime of crawls is dominated by starting/restarting Chromium. Making it possible to run with a single instance should be a pretty big perf win.

tidoust:
Starting with v14 (possibly with v12 as well?), Node.js supports loading ES6 modules within CommonJS modules dynamically through the `import()` method. This makes is possible to get rid of the `canonicalize-url.js` build, provided that the file gets renamed with a `.mjs` extension. This is what this update does.

This update also renames other ES6 modules with a `.mjs` extension. That is not really needed, but that seems better for consistency.

With one build that remained, it seemed a good idea to try to get rid of it as well. This is both harder and easier. Easier because `browser.js` can easily be generated dynamically through the Rollup API. Harder because that should be done only once when the program starts, which means that a setup phase needed to be created. Not fantastic, but hard to avoid.

In turn, that setup phase (along with teardown phase) allows to create the Puppeteer instance at the beginning and to reuse it throughout the crawl. Which this update also does.

A new test was added to make sure that building the code is doable. Other tests were fixed to call setup and teardown functions when needed.

This update means that Node.js >=14 is now needed. As such, the PR should only be merged once we have updated webref actions to also use v14.

Close #387.
Close #415.


dontcallmedom (commented):
Instead of calling setupBrowser and teardownBrowser from each CLI script, couldn't that be done from processSpecification (with a check that there isn't an existing browser instance to avoid launching several ones when doing batch processing?)

tidoust:
I started with that approach and dropped it because that didn't look "clean enough":

1. A teardown phase would still be needed, since `processSpecification` cannot guess when it is last called. I didn't like that asymetry. Granted, the teardown phase is not as critical as the setup phase, ending the program will close the browser.
2. That seemed like the kind of side effects that a function should not have. First time the function gets called, it would spend something like 500ms-1s initializing things.
3. Moreover, processSpecification is called in parallel. It's easy to unparallel things with a `Promise` but having to add that sync point shows that we'd be doing something clumsy.

A "cleaner" (as in "more functional") alternative would be to make setupBrowser return the variables instead of setting them globally, and passing them as arguments to processSpecification. But that means passing around a `browser` instance throughout the code :(

Another possibility would be to replace `processSpecification` with a `Processor` object. The constructor would call `setup` and the object would expose a `process` and `close` methods. The asymetry would remain but at least that would encapsulate the setup and teardown methods (instead of requiring external scripts to load them from `util`). We'd still need to pass that `Processor` instance around though. What about it?

dontcallmedom:
given the scope of this pull request, I think we should keep it simple - can we add another util function `processSingleSpec` that does the setup and teardown, so that we DRY the CLI scripts that process a single URL?

src/lib/util.js
```diff
@@ -35,6 +36,78 @@ function requireFromWorkingDirectory(filename) {
...
 
 
+/**
+ * Bundle the JS code that creates the reffy namespace in a browser context
+ * out of the source code in src/browserlib.
+ *
+ * @function
+ * @public
+ */
+async function buildBrowserlib() {
```
dontcallmedom, line 46: thinking some more about this - is there any reason to use rollup at this point? can't we just insert reffy.js as a module and let the browser do the imports?

tidoust: The same reason as initially ;) I've been using rollup because letting the browser do the imports is not that easy in practice: we want to inject a local script, but imports are going to be resolved against the base URL of the specification. For instance, if you're processing `https://www.w3.org/TR/presentation-api/`, then injecting `reffy.mjs` will trigger requests on `https://www.w3.org/TR/presentation-api/get-title.mjs` because `reffy.mjs` starts with:

```js
import getTitle from './get-title.mjs';
```

Now, the thing is we already intercept network requests, so we can intercept requests on script files and respond with the content of the local files instead. It's a bit convoluted (and cannot be very clean because we cannot use a different domain either, otherwise we'll run into CORS issues), but that seems to work. I'll push something. I'm glad I called the PR branch `shaving-yaks`.

tidoust:
> given the scope of this pull request, I think we should keep it simple - can we add another util function `processSingleSpec` that does the setup and teardown, so that we DRY the CLI scripts that process a single URL?

I hadn't done that on purpose because CLI scripts can also be used as modules, and calling scripts may be willing to call processing functions on a bunch of specs. That said, if we don't need to call rollup anymore, this situation would just be the same as before, which seems ok.

tests/extract-css.js
```diff
@@ -125,22 +126,30 @@ describe("Test CSS properties extraction", function() {
   this.slow(5000);
   this.timeout(10000);
   let browser;
+  let extractCSSCode;
   before(async () => {
+    // Convert the JS module to a JS script that can be loaded in Puppeteer
+    // without having to provide a URL for it (tests run in "about:blank" pages)
+    const bundle = await rollup.rollup({
```
dontcallmedom, line 133: this may be obvious, but I don't understand why we need rollup here?

dontcallmedom: OK, having tried various approaches, I now understand that there isn't a simple way to plug our module in a reusable context in puppeteer - this sounds like a gap in the platform (in particular, there is no way to import functions from an in-line `<script type=module></script>`), but that's for a different yak shaving day.

dontcallmedom (approved):

PR #487, opened 2021-01-27
Look for node_modules path when tool starts

### LINKED ISSUE #477: Using Reffy as a dependency is no longer possible due to #458
Using reffy as a dependency is no longer possible due to the changes in #458.  This is due to WebIDL2 being imported using a path relative to util.js.

https://github.com/w3c/reffy/blob/ab00f1bc152fdbb07984ffbc10cdbb75627d3b25/src/lib/util.js#L190-L192

When installed as a dependency, npm flattens module deps so that they are in the parent `node_modules`.

Note, this isn't a huge issue for me as I planned on moving away from reffy in favor of webref anyway - but I thought you should be aware in the event that you want this project to remain usable as a dependency.

tidoust:
The location of the `node_modules` folder depends on whether the code is run directly or as a dependency. When used as a dependency, all dependencies are flattened to the same parent `node_modules` folder.

The code needs to resolve relative links to `node_modules` because the `extract-dfns` browser library needs to import the WebIDL parser lib, which is installed as a dependency.

This update starts by setting a global `node_modules` constant, and uses it during spec processing to resolve the relative paths appropriately.

Fixes #477.


dontcallmedom (commented):
any way to add a test to check that reffy-as-a-module works in general?

tidoust:
I'm thinking a minimal test could:

1. Go to an empty folder and run `npm install <folder>` to install Reffy as a module
2. Run the same tests as in `tests/crawl.js` but against the `crawl-specs.js` file in the installed module to validate that crawling is going to work. That would typically have caught this issue.

In theory, it would be useful to test that the other tools exposed by the module run as intended, but that can probably be done over time...

What about it?

dontcallmedom:
Sounds like a good plan, thanks!

dontcallmedom (approved):

PR #500, opened 2021-02-04
Don't clobber previously identified definitions

### LINKED ISSUE #496: extractValueSpaces clobbers previously identified values
I noticed that when parsing [css-shapes-2](https://drafts.csswg.org/css-shapes-2/#typedef-shape-command), reffy fails to extract the syntax for several valuespaces.  After reviewing the code in [extract-cssdfn.mjs](https://github.com/w3c/reffy/blob/master/src/browserlib/extract-cssdfn.mjs#L141), it appears that reffy is capable of identifying these values.  The issue however, is that any later matches of the same syntax name will clobber any previously identified value.

This is true in both parseProductionRules:
https://github.com/w3c/reffy/blob/eba2d3069851bbf8b6d8215d6eddc0c296260ef2/src/browserlib/extract-cssdfn.mjs#L134-L136

And the final parsing rule:
https://github.com/w3c/reffy/blob/eba2d3069851bbf8b6d8215d6eddc0c296260ef2/src/browserlib/extract-cssdfn.mjs#L227-L235

I'm not sure if this is intended or not but instead of clobbering existing rules, reffy should consider only storing new results.  For example:

```javascript
if (!space.name in res) {
  res[space.name] = {};
}

if (space.prose && !res[space.name].prose) {
  res[space.name].prose = space.prose;
}
if (space.value && !res[space.name].value) {
  res[space.name].value = space.value;
}
```

pyoor:
Fixes #496

### DIFF
--- a/src/browserlib/extract-cssdfn.mjs
+++ b/src/browserlib/extract-cssdfn.mjs
@@ -131,8 +131,11 @@ const extractValueSpaces = doc => {
       .map(line => line.split(/\s?=\s/).map(s => s.trim().replace(/\s+/g, ' ')))
       .filter(val => val[0].match(/^<.*>$|^.*\(\)$/))
       .filter(val => !!val[1])
-      .forEach(val => res[val[0].replace(/^(.*\(\))$/, '<$1>')] = {
-        value: val[1]
+      .forEach(val => {
+        const name = val[0].replace(/^(.*\(\))$/, '<$1>')
+        if (!(name in res)) {
+          res[name] = { value: val[1] };
+        }
       });
 
   // Extract non-terminal value spaces defined in `pre` tags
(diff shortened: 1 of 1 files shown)


src/browserlib/extract-cssdfn.mjs
```diff
@@ -225,11 +228,14 @@ const extractValueSpaces = doc => {
     })
     .filter(space => !!space)
     .forEach(space => {
-      res[space.name] = {};
-      if (space.prose) {
+      if (!(space.name) in res) {
```
tidoust, line 231:
```suggestion
      if (!(space.name in res)) {
```

tidoust (approved):
Looks good (I'll commit the fix to the parenthesis position).

In an ideal world, we wouldn't be parsing definitions more than once. We do in practice:
1. within the same call to `parseProductionRules` (e.g. because we actually parse the parent of the extracted definition, and several definitions may have the same parent).
2. across calls to `parseProductionRules` (because the selectors overlap);
3. through the last dt parse that also overlaps with previous selectors.

This does not have real consequences in practice, other than the clobbering issue that this PR fixes, it just does not feel entirely satisfactory...

PR #568, opened 2021-04-14
Fix name for undefined.idl files

### LINKED ISSUE #537: There should be no IDL file named "undefined.idl"
Once in a while, it seems possible to end up with an IDL file named `undefined.idl`, e.g.:
https://github.com/w3c/webref/blob/09ab6914154ad6eaa1c3b589a88c12c940dcb8e3/ed/idlnames/undefined.idl

That should never happen.

tidoust:
The code could produce files named `undefined.idl` in the `idlnames` folder when it bumped on a partial definition for which the base definition does not exist in the crawl.

Ideally, these situations should never happen. They do in practice when a spec cannot be crawled e.g. for transient network issues or because it contains invalid IDL.

Note that the code was already correct for the `idlnamesparsed` folder.

Fixes #537.

### DIFF
--- a/src/cli/generate-idlnames.js
+++ b/src/cli/generate-idlnames.js
@@ -339,7 +339,7 @@ async function saveIdlNamesFragments(names, folder) {
   }
 
   await createFolderIfNeeded(folder);
-  await Promise.all(Object.values(names).map(idl => {
+  await Promise.all(Object.entries(names).map(([name, idl]) => {
     const res = [];
     if (idl.defined) {
       res.push(serializeNode(idl.defined));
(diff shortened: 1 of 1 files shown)


src/cli/generate-idlnames.js
```diff
@@ -339,15 +339,15 @@ async function saveIdlNamesFragments(names, folder) {
...
+  await Promise.all(Object.entries(names).map(([name, idl]) => {
     const res = [];
     if (idl.defined) {
       res.push(serializeNode(idl.defined));
     }
     if (idl.extended) {
       idl.extended.map(node => res.push(serializeNode(node)));
     }
-    const filename = path.join(folder, idl.name + '.idl');
+    const filename = path.join(folder, name + '.idl');
```
dontcallmedom, line 350: I'm not sure I understand how this works - why won't undefined no longer appear?

tidoust: The previous code assumed that the `name` property was always set. That is not the case when the object only contains partial definitions. FYI, the code that generates these objects is at:
https://github.com/w3c/reffy/blob/master/src/cli/generate-idlnames.js#L158.

The fix is to use the key to the object in `names` instead, which is always set to the interface name.

Same logic as the one used for the `idlnamesparsed` folder:
https://github.com/w3c/reffy/blob/master/src/cli/generate-idlnames.js#L316

dontcallmedom (approved):

PR #592, opened 2021-05-11
When reffy.mjs has an error, log it.

jyasskin:
Without this, it was hard to debug mistakes in `reffy.mjs` and its dependencies. There could be better ways to do it.


tidoust (approved):
Thanks! It would also make sense to report such errors somehow. That is, in regular crawls, such errors could also be triggered by broken scripts in specs, which would be good things to detect and report. I created #594 to track that.

PR #611, opened 2021-05-28
Add generated IDL attribute names to CSS properties

tidoust:
See related discussion in:
https://github.com/w3c/webref/issues/51

This update adds a `styleDeclaration` property to CSS property definitions in CSS extracts. The property contains the list of IDL attribute names that the CSS property generates per the CSSOM spec on the `CSSStyleDeclaration` interface:
https://drafts.csswg.org/cssom/#ref-for-css-property-to-idl-attribute

The implementation of the CSS property to IDL attribute algorithm was copied over from:
https://github.com/foolip/mdn-bcd-collector/blob/c2126e8f367f08333a0f7b0604e42c8fdc11d182/build.js#L569

Note that the creation of this property is done after the crawl and not directly within the `extract-cssdfn.mjs` logic because that logic only extracts CSS properties that are defined in table, and the crawler completes that list afterwards with CSS properties that are defined in prose (such as those in the compat spec).

### DIFF
--- a/src/lib/util.js
+++ b/src/lib/util.js
@@ -584,6 +584,56 @@ async function expandCrawlResult(crawl, baseFolder) {
 }
 
 
+/**
+ * Retrieves the list of IDL attribute names that the CSS property generates
+ * per the CSSOM spec, see:
+ * https://drafts.csswg.org/cssom/#ref-for-css-property-to-idl-attribute
+ *
+ * @function
+ * @param {String} property CSS property name
+ * @return {Array(String)} An array of IDL attribute names
+ */
+function getGeneratedIDLNamesByCSSProperty(property) {
+    const res = [];
+
+    // Converts a CSS property to an IDL attribute name per the CSSOM spec:
+    // https://drafts.csswg.org/cssom/#css-property-to-idl-attribute
+    function cssPropertyToIDLAttribute(property, lowercaseFirst) {
+        let output = '';
+        let uppercaseNext = false;
+        if (lowercaseFirst) {
+            property = property.substr(1);
+        }
+        for (const c of property) {
+            if (c === '-') {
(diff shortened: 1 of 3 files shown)


tests/util.js
```diff
@@ -76,4 +79,25 @@ describe('isLatestLevelThatPasses', () => {
...
+      ['touch-action', 'touchAction']);
+  });
+
+  it('returns the camel-cased, webkit-cased and dashed attribute names for "-webkit-background-clip"', () => {
+    assert.deepEqual(
+      getGeneratedIDLNamesByCSSProperty('-webkit-background-clip'),
+      ['-webkit-background-clip', 'WebkitBackgroundClip', 'webkitBackgroundClip']);
+  });
+
+  it('returns the camel-cased attribute name for "display"', () => {
```
foolip, line 98:
```suggestion
  it('returns just the name for "display"', () => {
```

src/lib/util.js
```diff
@@ -584,6 +584,56 @@ async function expandCrawlResult(crawl, baseFolder) {
...
+    if (property.startsWith('-webkit-')) {
+        res.push(cssPropertyToIDLAttribute(property, true));
+    }
+
+    // Add dashed attribute if needed
+    if (property.includes('-')) {
+        res.push(property);
+    }
+
+    return res.sort();
```
foolip, line 633: It would be more convenient if the result was ordered in a predictable way, perhaps with the dashed (unmodified) attribute first, then camel-case if it's different, and finally the webkit special case if different. I'd then use the first entry for some tests.

foolip (approved):
Looks good, and I grant a license to the copied code under https://github.com/w3c/reffy/blob/master/LICENSE.

foolip (approved):

PR #615, opened 2021-05-31
Extract mapping between HTML/SVG elements and their interface

tidoust:
See discussion in:
https://github.com/w3c/webref/issues/120

This update makes the crawler parse element definitions in HTML and SVG specs, and create a JSON extract under an `elements` folder per spec.

In the JSON extract, each element is represented by an object with a `name` property that is the element's name, and an `interface` property that contains the IDL interface associated with the element.

If the need arises, the extraction code could be expanded to extract further information from the definitions tables. It does not extract anything else for now, because there are always exceptions-to-the-rule that make it hard to create a machine-readable dump (for instance, once in a while, there are specific conditions to describe the contexts under which an element may appear, and these specific conditions are explained in human-readable prose).


dontcallmedom (approved):
great addition, thanks!

PR #630, opened 2021-06-08
Remove code needed to annotate elements in HTML spec

dontcallmedom:
Done upstream in https://github.com/whatwg/html/pull/6750

(marked as draft until the PR is merged at which point I'll double check we get the same results in extracted definitions)


dontcallmedom:
HTML spec now updated and I've verified the resulting definitions match what we want

tidoust (approved):

PR #631, opened 2021-06-08
Remove code needed to annotate idl definitions in HTML spec

dontcallmedom:
Done upstream in https://github.com/whatwg/html/pull/5957

(marked as draft until the PR is merged at which point I'll double check we get the same results in extracted definitions)

### DIFF
--- a/src/browserlib/extract-dfns.mjs
+++ b/src/browserlib/extract-dfns.mjs
@@ -383,15 +383,6 @@ function preProcessHTML() {
       }
       let m;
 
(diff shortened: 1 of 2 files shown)


dontcallmedom:
https://github.com/whatwg/html/pull/5957 got merged, but re-running the extractions highlighted a few gaps, submitted in https://github.com/whatwg/html/pull/6751

dontcallmedom:
the diff is now empty, so should be good to go

PR #633, opened 2021-06-09
Fix elements extraction for /TR version of SVG 1.1

### LINKED ISSUE #632: Elements extraction crashes on /TR version of SVG 1.1
See https://github.com/w3c/webref/blob/master/tr/index.json#L14461

tidoust:
Fixes #632

### DIFF
--- a/src/browserlib/extract-elements.mjs
+++ b/src/browserlib/extract-elements.mjs
@@ -137,12 +137,12 @@ export default function () {
   // Extract SVG elements that use the "element-summary" pattern
   const svgSummaryElements = [...document.querySelectorAll('div.element-summary')]
     .map(el => {
-      const dfn = el.querySelector('.element-name dfn');
+      let dfn = el.querySelector('.element-name');
       if (!dfn) {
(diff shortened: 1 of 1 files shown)


src/browserlib/extract-elements.mjs
```diff
@@ -137,12 +137,12 @@ export default function () {
   // Extract SVG elements that use the "element-summary" pattern
   const svgSummaryElements = [...document.querySelectorAll('div.element-summary')]
     .map(el => {
-      const dfn = el.querySelector('.element-name dfn');
+      let dfn = el.querySelector('.element-name');
```
dontcallmedom, line 140: why s/const/let/ ?
it's a bit confusing to keep the variable name `dfn` if this is no longer targeting the `dfn` element?

tidoust: Ah, good point. I started the patch with an alternative, then realized I didn't need it and forgot to switch back to `const`.

dontcallmedom (approved):

PR #643, opened 2021-06-10
Remove code for annotating HTML spec definitions overtaken by upstream…

dontcallmedom:
… changes

not to be merged until https://github.com/whatwg/html/pull/6757 is, and before I can verify the resulting definitions aren't changed


dontcallmedom:
we should be good to go after whatwg/html#6776 gets merged

dontcallmedom:
in fact all the changes here are also in #646, so closing this

PR #656, opened 2021-06-17
Use absolute URLs for dfn headings, headings, and ids

### LINKED ISSUE #647: Multipage specs don't indicate pages in headings data, leading to collisions/corruption
In the headings data for a multipage spec, like SVG2 https://github.com/w3c/webref/blob/master/tr/headings/SVG2.json, there's no indication of what page each heading is on.

This makes the data useless for actually linking (most of the IDs don't exist at the url provided at the top of the data file, they're in sub-pages) and causes collisions (multiple "Overview" ids in that file, for example)
and data corruption (check the "id": "Overview" entries in the SVG2 file linked above - they both list their number as 12.1, despite the first clearly being intended to be 5.4.1).

Each entry should have an optional "path" member listing the additional path that needs to be concatenated to the top-level url to reach the page that the ID is actually on; if it's missing, it means the ID is on the top-level page (or it's a single-page spec).

tidoust:
Multipage specs were badly supported because the code assumed that local IDs would be unique for the entire document, but that is generally speaking not the case.

This update adds an `href` property with an absolute URL throughout extracts. That was already the case for dfns. This is now also done for headings in dfn extracts and for the headings extracts themselves.

The ids extracts now also use absolute URLs everywhere.

Local IDs still appear in the `id` property.

Fixes #647.

Instead of an absolute URL with an ID, an alternative would be to have a `page` property that contains an absolute URL to the page where the heading is defined (or a relative URL to the base URL of the spec). Net benefit is that the extracts would be smaller in size as putting absolute URLs everywhere basically doubles the size of the extracts. One drawback is that would then be up to the consumer of the data to assemble the absolute URL with fragment. Given that we already had an `href` property for the dfns, I thought it made sense to be consistent and use the same mechanism everywhere.

### DIFF
--- a/src/browserlib/extract-ids.mjs
+++ b/src/browserlib/extract-ids.mjs
@@ -1,12 +1,28 @@
 /**
- * Extract ids from documents
+ * Extract absolute ids from documents
 */
 export default function () {
-  return [...document.querySelectorAll('*[id]')].map(n => n.id).concat(
+  // Compute once whether we created a single page version out of multiple pages
+  const isMultipage = !!document.querySelector('[data-reffy-page]');
+
+  function getAbsoluteUrl(node) {
+    const page = isMultipage ?
+      node.closest('[data-reffy-page]')?.getAttribute('data-reffy-page') :
+      null;
+    const url = new URL(page ?? window.location.href);
+    url.hash = '#' + node.id;
+    return url.toString();
+  }
+
+  return [...document.querySelectorAll('*[id]')]
+    .map(n => getAbsoluteUrl(n))
+
     // Capture anchors set in <a name> when they're not dup of ids
-    [...document.querySelectorAll('a[name]')]
-      .filter(n => !n.id || n.id !== n.name).map(n => n.name)
-    // We ignore respec- prefixed ids to avoid keeping track of their evolution
+    .concat([...document.querySelectorAll('a[name]')]
+      .filter(n => !n.id || n.id !== n.name).map(n => getAbsoluteUrl(n))
+    )
+
+    // I respec- prefixed ids to avoid keeping track of their evolution
     // They're clearly not meant to be link target in any case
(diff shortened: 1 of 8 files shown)


src/browserlib/extract-ids.mjs
```diff
@@ -1,12 +1,28 @@
...
+
     // Capture anchors set in <a name> when they're not dup of ids
-    [...document.querySelectorAll('a[name]')]
-      .filter(n => !n.id || n.id !== n.name).map(n => n.name)
-    // We ignore respec- prefixed ids to avoid keeping track of their evolution
+    .concat([...document.querySelectorAll('a[name]')]
+      .filter(n => !n.id || n.id !== n.name).map(n => getAbsoluteUrl(n))
+    )
+
+    // I respec- prefixed ids to avoid keeping track of their evolution
```
dontcallmedom, line 25:
```suggestion
    // Ignore respec- prefixed ids to avoid keeping track of their evolution
```

src/browserlib/map-ids-to-headings.mjs
```diff
@@ -36,6 +36,18 @@ export default function () {
   const { outline, nodeToSection } = createOutline(document.body);
   const sections = flattenSections(outline);
 
+  // Compute once whether we created a single page version out of multiple pages
+  const isMultipage = !!document.querySelector('[data-reffy-page]');
+
+  function getAbsoluteUrl(node) {
+    const page = isMultipage ?
```
dontcallmedom, line 43: should this be factored out since that function is repeated 4 times?

dontcallmedom (approved):
LGTM with a couple of suggested improvements

src/browserlib/map-ids-to-headings.mjs
tidoust: Done.

PR #659, opened 2021-06-21
Ignore dfns that have an invalid type

### LINKED ISSUE #658: Definitions should be verified to have a valid dfn type
While reviewing the effects of switching Bikeshed's dfn harvester to Reffy, I noticed a new dfn cropped in with a type of "idl" <https://github.com/w3c/webref/blob/master/tr/dfns/payment-request.json#L1992>

This isn't a valid [definition type](https://tabatkins.github.io/bikeshed/#dfn-types). (It's a [linking supertype](https://tabatkins.github.io/bikeshed/#link-types) that encompasses all the IDL-related types.)

I guess the type of a dfn should be verified to be valid, and the dfn omitted if it's not?

(I [reported this on ReSpec as well](https://github.com/w3c/respec/issues/3644), since it's not being reported as an error in the source document, but regardless of the inputs, the Reffy data should be clean.)

tidoust:
This update restricts dfns extraction to definitions that have a valid type, as defined in:
https://tabatkins.github.io/bikeshed/#dfn-types
(+ `namespace` and `event` that are valid but do not yet appear in the doc)

In practice, the only invalid type that appears in Editor's Drafts is `idl`, which ReSpec uses to flag internal slots. To avoid losing definitions while ReSpec gets fixed and while the changes propagate to all specs and /TR versions of the spec, the code automatically converts `idl` definition types to `attribute`.

Data validation is something that we would typically do at a later stage, through anomaly reports, or through patches as we produce packages. That said, dfns are more time sensitive than other types of data that Reffy extracts from the specs and the validation process would need to be semi-manual.

Dfns with an invalid type get reported as warnings to the console.

Fixes #658.

### DIFF
--- a/src/browserlib/extract-dfns.mjs
+++ b/src/browserlib/extract-dfns.mjs
@@ -29,11 +29,77 @@ import {parse} from "../../node_modules/webidl2/index.js";
  * @return {Array(Object)} An Array of definitions
 */
 
-function definitionMapper(el, idToHeading) {
-  function normalize(str) {
-    return str.trim().replace(/\s+/g, ' ');
-  }
+function normalize(str) {
+  return str.trim().replace(/\s+/g, ' ');
+}
 
+// Valid types defined in https://tabatkins.github.io/bikeshed/#dfn-types
+// (+ "namespace" and "event" which are not yet in the doc)
+function hasValidType(el) {
+  const validDfnTypes = [
+    // CSS types
+    'property',
+    'descriptor',
+    'value',
+    'type',
+    'at-rule',
+    'function',
+    'selector',
+
+    // Web IDL types
+    'namespace',
+    'interface',
+    'constructor',
+    'method',
+    'argument',
+    'attribute',
+    'callback',
+    'dictionary',
+    'dict-member',
+    'enum',
+    'enum-value',
+    'exception',
+    'const',
+    'typedef',
+    'stringifier',
+    'serializer',
+    'iterator',
+    'maplike',
+    'setlike',
+    'extended-attribute',
+    'event',
+
+    // Element types
+    'element',
+    'element-state',
+    'element-attr',
+    'attr-value',
+
+
+    // URL scheme
+    'scheme',
+
+    // HTTP header
+    'http-header',
+
+    // Grammar type
+    'grammar',
+
+    // "English" terms
+    'abstract-op',
+    'dfn'
+  ];
+
+  const type = el.getAttribute('data-dfn-type') ?? 'dfn';
+  const isValid = validDfnTypes.includes(type);
+  if (!isValid) {
+    console.warn('[reffy]', `"${type}" is an invalid dfn type for "${normalize(el.textContent)}"`);
+  }
+  return isValid;
+}
+
+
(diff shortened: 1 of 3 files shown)


src/browserlib/extract-dfns.mjs
```diff
@@ -140,6 +206,18 @@ export default function (spec, idToHeading = {}) {
   }
 
   return [...document.querySelectorAll(definitionsSelector)]
+    .map(node => {
+      // 2021-06-21: Temporary preprocessing of invalid "idl" dfn type (used for
+      // internal slots) while fix for https://github.com/w3c/respec/issues/3644
+      // propagates to all EDs and /TR specs. To be dropped once crawls no
+      // longer produce warnings.
+      if (node.getAttribute('data-dfn-type') === 'idl') {
+        node.setAttribute('data-dfn-type', 'attribute');
```
dontcallmedom, line 215: I've commented in https://github.com/w3c/respec/issues/3644#issuecomment-864894696 - I'm not convinced `attribute` is the right approach.

Code looks good otherwise.

src/browserlib/extract-dfns.mjs
```diff
@@ -140,6 +206,18 @@ export default function (spec, idToHeading = {}) {
...
+        console.warn('[reffy]', `Fixed invalid "idl" dfn type "${normalize(node.textContent)}"`);
+      }
+      return node;
+    })
+    .filter(hasValidType)
```
dontcallmedom, line 220: Thinking some more - instead of removing definitions with an invalid type, shouldn't we default them back to `dfn`?

tidoust: I'm hesitant to do that. The good thing about removing definitions with an invalid type is that no one can reference them. When someone needs to do that, they will notice that the dfn is missing and raise an issue somewhere to have that fixed, and this should eventually appear on our radar.

If we default these definitions to `dfn`, people may just ask for a `data-export` attribute to be added, and then they can start to link to it, probably using a wrong shorthand syntac. We may not notice the issue until both the source spec that defines the term and the spec that references it are wrong.

Obviously, we could notice if we had a proper way to deal with the "warnings" that Reffy logs. I'm not utterly confident that we'll look into the workflow logs, and it's more work to have a better alert mechanism. Useful work but more work... Plus, if we do that, we can just as well remove the dfns: we would have them fixed in the source spec.

dontcallmedom: ok, I can see that making kind of sense; the risks are probably limited in any case if [ReSpec starts flagging unknown definition types](https://github.com/w3c/respec/pull/3646#issuecomment-865003774) the same way bikeshed does.

src/browserlib/extract-dfns.mjs
tidoust: We're still discussing whether it would be worth creating an additional type in speced/spec-dfn-contract#1.

In the meantime, w3c/respec#3644 was closed with ReSpec now flagging internal slots with an `attribute` type and internal methods with a `method` type. I've updated the code to the same here and propose to go ahead with this while we converge on new types (or on making these dfns private by default one way or another).

dontcallmedom (approved):

PR #664, opened 2021-06-23
Merge CSS property definitions when possible

### LINKED ISSUE #662: Support multiple partial definitions of a CSS property
The /TR version of GCPM Level 3 contains two partial definitions for the `content` property:
- https://www.w3.org/TR/css-gcpm-3/#element-syntax
- https://www.w3.org/TR/css-gcpm-3/#leaders

Reffy does not know how to combine these definitions into and now throws an exception.

Via https://github.com/w3c/webref/pull/284#pullrequestreview-690294560

### LINKED ISSUE #663: Exclude CSS property definitions that describe "changes"
The /TR version of CSS 2.1 contains a Change section that describes [changes made to the definition of the `top` property](https://www.w3.org/TR/CSS21/changes.html#q21.523). Reffy picks up that change definition as a real definition and throws an exception because that property is already defined elsewhere.

There is no immediate markup that can be used to exclude the definition. Perhaps the most generic method would be to exclude definitions that have `<ins>` and `<del>` tags within the definition.

Via https://github.com/w3c/webref/pull/284#pullrequestreview-690294560

tidoust:
This makes the crawler merge CSS property definitions when it finds multiple definitions for a CSS property in the same spec. The merge operation remains basic: it merely merges the new values with the initial ones and hardly checks anything else. Fixes #662.

The update also makes Reffy ignore definitions that contain `<ins>` or `<del>` tags. Such definitions should only appear in "Changes" sections to detail changes that were made to the initial definition over time. Fixes #663.

### DIFF
--- a/src/browserlib/extract-cssdfn.mjs
+++ b/src/browserlib/extract-cssdfn.mjs
@@ -89,12 +89,57 @@ const extractDlDfn = dl => {
 
 
 /**
+ * Merges CSS definitions for the same property into one
+ *
+ * The function runs a few simple sanity checks on the definitions. More checks
+ * would be needed to fully validate that merging is fine.
+ *
+ * The function returns null if CSS definitions cannot be merged.
+ */
+const mergeDfns = (dfn1, dfn2) => {
+  // Definitions should be about the same property
+  if (dfn1.name !== dfn2.name) {
+    return null;
+  }
+
+  // There should never be two base definitions for the same CSS property
+  if (dfn1.value && dfn2.value) {
+    return null;
+  }
+
+  const baseDfn = dfn2.value ? dfn2 : dfn1;
+  if (!baseDfn.value && !baseDfn.newValues) {
+    return null;
+  }
+
+  const partialDfn = (baseDfn === dfn1) ? dfn2 : dfn1;
+  if (!partialDfn.newValues || partialDfn.initial) {
+    return null;
+  }
+  if (partialDfn.initial && partialDfn.initial !== baseDfn.initial) {
+    return null;
+  }
+
+  const merged = Object.assign(baseDfn);
+  if (merged.value) {
+    merged.value += ` | ${partialDfn.newValues}`;
+  }
+  else {
+    merged.newValues += ` | ${partialDfn.newValues}`;
+  }
+
+  return merged;
+};
+
+
+/**
  * Extract CSS definitions in a spec using the given CSS selector and extractor
  */
 const extractDfns = (doc, selector, extractor, { unique } = { unique: true }) => {
   let res = {};
   [...doc.querySelectorAll(selector)]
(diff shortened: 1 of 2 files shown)


src/browserlib/extract-cssdfn.mjs
```diff
@@ -88,13 +88,58 @@ const extractDlDfn = dl => {
...
+  if (dfn1.name !== dfn2.name) {
+    return null;
+  }
+
+  // There should never be two base definitions for the same CSS property
+  if (dfn1.value && dfn2.value) {
+    return null;
+  }
+
+  const baseDfn = dfn2.value ? dfn2 : dfn1;
```
dontcallmedom, line 110: Maybe combine the assignments of base & partial to help with readability?
```suggestion
  const [baseDfn, partialDfn] = dfn2.value ? [dfn2, dfn1] : [dfn1, dfn2];
```

src/browserlib/extract-cssdfn.mjs
```diff
@@ -88,13 +88,58 @@ const extractDlDfn = dl => {
...
+  if (!baseDfn.value && !baseDfn.newValues) {
+    return null;
+  }
+
+  const partialDfn = (baseDfn === dfn1) ? dfn2 : dfn1;
+  if (!partialDfn.newValues || partialDfn.initial) {
+    return null;
+  }
+  if (partialDfn.initial && partialDfn.initial !== baseDfn.initial) {
```
dontcallmedom, line 119: `partialDfn.initial` can't be true-ish at this stage given the previous condition, can it?

src/browserlib/extract-cssdfn.mjs
```diff
@@ -103,7 +148,11 @@ const extractDfns = (doc, selector, extractor, { unique } = { unique: true }) =>
     .forEach(dfn => {
       if (res[dfn.name]) {
         if (unique) {
-          throw new Error(`More than one dfn found for CSS property/descriptor "${dfn.name}"`);
+          const merged = mergeDfns(res[dfn.name], dfn);
+          if (!merged) {
+            throw new Error(`More than one dfn found for CSS property "${dfn.name}" and dfns cannot be merged`);
```
dontcallmedom, line 153: just to be sure, we don't expect any current spec to throw, right?

src/browserlib/extract-cssdfn.mjs
tidoust: Ooops. Previous condition shouldn't have been there. Removed.

src/browserlib/extract-cssdfn.mjs
tidoust: The update is motivated by two current specs for which previous code throws. Once that is merged, all current specs should pass.

dontcallmedom (approved):

PR #697, opened 2021-07-22
Adjust CSS dfns extraction rules

tidoust:
CSS Animations now uses `<span class=prod>` instead of `<pre class=prod>` for production rules (see https://github.com/w3c/webref/pull/307#pullrequestreview-712521622). The update extends the query selector accordingly (we could probably just use `.prod` but that seems slightly too generic).


tidoust:
Taking the libery to merge. I made sure that this change does not introduce false positives: it merely brings back the value spaces that should not have disappeared in css-animations.

PR #710, opened 2021-08-05
Structure anomaly report by Working Group

dontcallmedom:
This helps going through the list of one's relevant issues

### DIFF
--- a/src/cli/generate-report.js
+++ b/src/cli/generate-report.js
@@ -29,7 +29,6 @@ const fetch = require('../lib/util').fetch;
  */
 const byTitle = (a, b) => a.title.toUpperCase().localeCompare(b.title.toUpperCase());
 
-
 /**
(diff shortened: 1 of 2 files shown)


tidoust (commented):
This currently breaks the JavaScript filtering logic in the HTML version of the anomaly file that eventually gets produced from the Markdown version, meaning the one under https://w3c.github.io/webref/ed/

Typically, that filtering logic expects specs' titles to be in `<h2>` (see https://github.com/w3c/webref/blob/master/common/reffy.js#L19), whereas this update creates `<h3>` for them.

More generally speaking, the filtering and section expanding/collapsing logic probably needs to be updated?

Of course, the question is: is it worth updating that logic or should we take that opportunity to create a more directly useful anomaly report?

tidoust (approved):

PR #753, opened 2021-09-30
Accept series shortnames as input for specs

tidoust:
If a shortname is not a spec shortname, Reffy now looks for a match in series shortname, and crawls the current specification in the series.


dontcallmedom (approved):
you couldn't resist, could you >:)

tidoust:
Too tempting :))

PR #756, opened 2021-10-01
Create cache folder in install folder, not run folder

tidoust:
The local cache folder (`.cache`) was created where the command was run. If you install Reffy globally, you won't expect it to create files in the folder you run it into by default.

This update creates the `.cache` folder in the install folder by default. It may not be good practice to put a cache that will typically contain dozens of MB of data in there, but that seems better than putting a cache folder where ever the command is run in any case.

Note one can override the default through the `cacheFolder` setting in a `config.json` file.

### DIFF
--- a/src/lib/fetch.js
+++ b/src/lib/fetch.js
@@ -43,6 +43,11 @@ async function fetch(url, options) {
         options.refresh = 'once';
     }
 
+    // Use cache folder where Reffy was installed, not where it is run
+    if (!options.cacheFolder) {
+        options.cacheFolder = path.resolve(__dirname, '..', '..', '.cache');
+    }
+
     return baseFetch(url, options);
 }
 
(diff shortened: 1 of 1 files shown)


dontcallmedom:
hmm... At least on linux, I think the expectation would be that a global install would have to be run with higher privileges and will put the script in an otherwise-readonly place for regular users, where in particular creating a cache folder at run time would fail.

Maybe we should use something à la https://stackoverflow.com/a/31999308/139591

tidoust:
Ah, right. I updated the PR to use a folder in the OS temp folder, which is way cleaner in any case. As opposed to the question on Stack Overflow, we don't need the cache to be shared across users for instance, so we can simplify the logic a bit.

dontcallmedom (approved):

Issue #786, opened 2021-11-09
CI tests may keep running for 3 hours for no apparent reason


tidoust:
This may be due to the nock proxy not receiving the requests that it expects to receive, but CI tests tend to run forever recently when run as an action. I cannot reproduce that locally though.

tidoust:
Main reason is likely that tests created multiple instances of Puppeteer at once, and did not reclaim them properly. Now fixed.
A couple of improvements were also made to the mock server used in the tests to intercept network requests.

I will optimistically close the issue. To be re-opened if the fix and improvements did not actually solve the issue...

PR #812, opened 2022-01-03
Improve network interception test logic

tidoust:
Use of the Nock server could easily create situations where an exception would be thrown outside of Mocha's scope and would not be correctly reported. The code no longer throws errors from event listeners.

Also, by default, the crawler actually falls back to Puppeteer's network stack when network interception fails for some reason. That network stack is out of reach for the Nock server. The code now sports a `forceLocalFetch` crawl option to disable that fallback and disabled the cache (although I note that I'm not yet fully convinced that disabling the cache actually works for all resources).

A couple of request interceptions seem no longer needed and get removed in that update too.

Also, there was no proper test on whether the nock server received all expected requests (call to `isDone()` is a good start but that only returns a boolean). The code now throws when expected requests are missing.

The above updates are made in the hope that they would fix CI tests that fail after 6 hours (#786), but then none of them should really explain why these timeouts occurred.


tidoust:
I believe I found a possible explanation for the timeouts: tests could create Puppeteer instances that were never closed (see https://github.com/w3c/reffy/pull/812/commits/c470f9784e6564e8d1e931991097485357107685)

dontcallmedom (approved):

PR #814, opened 2022-01-06
Add parsed IDL generator tool

tidoust:
The tool contains the logic to generate parsed IDL structures from the raw IDL. It takes a crawl report as input and creates JSON files in an `idlparsed` folder.

The crawler now uses the functions exported by the tool to generate the `idlparsed` extracts.

The tool is needed to creates curated versions of reports and in particular to re-generate the `idlparsed` extracts once IDL patches have been applied to the crawled IDL.

This update includes some tests on the generation.

### DIFF
--- a/src/cli/generate-idlparsed.js
+++ b/src/cli/generate-idlparsed.js
@@ -28,7 +28,8 @@ const path = require('path');
 const webidlParser = require('../cli/parse-webidl');
 const {
   expandCrawlResult,
-  requireFromWorkingDirectory
+  requireFromWorkingDirectory,
+  createFolderIfNeeded
 } = require('../lib/util');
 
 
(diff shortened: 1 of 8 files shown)


src/cli/generate-idlparsed.js
```diff
@@ -0,0 +1,155 @@
...
+
+async function generateIdlParsedFromPath(crawlPath) {
+  const crawlIndex = requireFromWorkingDirectory(path.resolve(crawlPath, 'index.json'));
+  const crawlResults = await expandCrawlResult(crawlIndex, crawlPath, ['idl']);
+  await Promise.all(crawlResults.results.map(generateIdlParsed));
+  return crawlResults;
+}
+
+
+async function createFolderIfNeeded(name) {
```
dontcallmedom, line 76: we have this [in generate-idlnames.js](https://github.com/w3c/reffy/blob/ef75e94483e8238b9c6879a5d48b9e60942c5dac/src/cli/generate-idlnames.js#L311) - it could be factored out in util.js?

src/cli/generate-idlparsed.js
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  const subfolder = path.join(folder, 'idlparsed');
+  await createFolderIfNeeded(subfolder);
+
+  if (!spec?.idl?.idl) {
+    return;
+  }
+
+  delete spec.idl.idl;
```
dontcallmedom, line 117: wouldn't it be cleaner to delete it only in the generated object? it feels a bit foot-gunny to modify the original object in a function called "save"

tidoust: Yes, that requires making at least a shallow copy of the object, which the code currently does not do.

The cleanest option would be to stop creating a hybrid structure under `idl` that holds both the result of parsing the raw IDL *and* the raw IDL (under an `idl` property), especially since that's clumsy and we need to separate them in the rest of the code in any case. The parsed IDL should rather be stored in a separate `idlparsed` property. It is a breaking change for code that use Reffy to do further processing on crawl results (e.g. Strudy), but it's easy to update code that relies on this, so this update seems a good opportunity to shave that yak...

src/cli/generate-idlparsed.js
tidoust: Done. Further factoring of the save functions could be envisioned but well, it ain't broken...

dontcallmedom (approved):

PR #825, opened 2022-01-21
Add option to use fallback data when crawling spec list

dontcallmedom:
not tested

I'm thinking this approach might help reducing risks for bogus updates to webref

### DIFF
--- a/reffy.js
+++ b/reffy.js
@@ -149,6 +149,12 @@ Usage notes for some of the options:
   Provides an existing JSON crawl data file to use as a source of fallback data
   for specs that fail to be crawled.
 
+  The fallback data gets copied as-is. It is the responsibility of the caller
+  to make sure that extracts it may link to actually exist and match the ones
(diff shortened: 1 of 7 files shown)


tidoust (approved):
@dontcallmedom I've made some changes to the PR. Could you review?

Changes are to:
- handle `file://` URLs correctly (they yield a 0 status code when loaded with Puppeteer)
- detect and report as errors CSS server issues as well
- move the fallback logic to the main crawl function
- match on the initial URL instead of on the shortname
- preserve the `error` property when fallback data is used, to be able to tell that fallback data was used (and understand why it was used)

dontcallmedom:
I can't github-approve my own PR apparently, but I reviewed your fixes and improvements and they LGTM

tidoust:
Well, my fixes didn't really LGTM, so I had another look ;)

The crawl would lose itself in extracts and produce garbage. I integrated spec result expansion in the mix to make sure that the crawl function returns the right result, and that the save function generates the right exports.

PR #826, opened 2022-01-21
Expose idlnames and idlparsed utility functions in package

tidoust:
This makes it easier to reuse these functions from within other packages. Typically useful to create a curated data view in Webref.

The update also adjusts the `expandCrawlResults` function to convert CRLF to LF when it loads text files, as Git may auto-convert LF to CRLF on checkout on Windows machines (that does not create any error per se, but we'll want to reuse the expanded IDL to create multiline IDL fragments in parsed IDL constructs, and it seems good to make sure that all platforms generate the exact same IDL fragments).

### DIFF
--- a/src/lib/util.js
+++ b/src/lib/util.js
@@ -740,6 +740,13 @@ async function expandCrawlResult(crawl, baseFolder, properties) {
                 const filename = path.join(baseFolder, spec[property]);
                 contents = await fs.readFile(filename, 'utf8');
             }
+
+            // Force Linux-style line endings
+            // (Git may auto-convert LF to CRLF on Windows machines and we
+            // want to store multiline IDL fragments as values of properties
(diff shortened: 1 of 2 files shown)


src/lib/util.js
```diff
@@ -740,6 +740,13 @@ async function expandCrawlResult(crawl, baseFolder, properties) {
                 const filename = path.join(baseFolder, spec[property]);
                 contents = await fs.readFile(filename, 'utf8');
             }
+
+            // Force Linux-style line endings
```
dontcallmedom, line 744:
```suggestion
            // Force UNIX-style line endings
```

tidoust:
I had not realized that `replaceAll` got added to Node.js starting with v15 (GitHub action runs v14). I'll update the PR.

dontcallmedom (approved):

Issue #854, opened 2022-02-03
Fallback mechanism introduces final new line in extracts


tidoust:
When used, the fallback mechanism introduces a final new line in the extracts produced (at least in the IDL extracts). That does not change the IDL meaning but:

1. It creates an artificial and temporary update to the extract in the repo
2. If a patch applies to the extract, the patch may not apply to the new extract because of the new line.

This is likely due to the `expandSpecResult` function.

A recent example: https://github.com/w3c/webref/commit/e95adca52e22ab9e57ffcee582c4423a0cca44e4#diff-cc6be26e89418a24d8443d1395cd3596257ad0616f6f3baf046322e899ee2af8

tidoust:
Fixed by #855

PR #855, opened 2022-02-03
Remove trailing line return in IDL when loading from fallback

dontcallmedom:
Otherwise new line returns gets added when the fallback is used; among other things, this makes webref curated patches no longer applicable as in https://github.com/w3c/webref/runs/5051707336?check_suite_focus=true


tidoust (approved):
Thanks! This fixes #854.

PR #856, opened 2022-02-04
Record and use HTTP cache information to skip unnecessary crawling and processing

### LINKED ISSUE #850: Make reffy crawler http-cache friendly
An average reffy crawl takes around 6 to 8 minutes, and that duration is set to expand as we consider [expanding the source of crawled specifications](https://github.com/w3c/browser-specs/issues/436).

But in practice, the vast majority of the time is spent fetching and processing documents that have not changed since the last crawl.

If we were to record the ETag and Last-Modified header our crawler hits for each crawled URLs, we could easily skip the large majority of the work using the fallback path developed in #825 when hitting a 304.

dontcallmedom:
close #850

### DIFF
--- a/src/lib/specs-crawler.js
+++ b/src/lib/specs-crawler.js
@@ -103,7 +103,8 @@ async function crawlSpec(spec, crawlOptions) {
               etag, lastModified}
         );
         if (result.error === "reuseexistingdata" && fallback) {
-          crawlOptions.quiet ?? console.warn(`skipping ${spec.url}, no change`);          const copy = Object.assign({}, fallback);
+          crawlOptions.quiet ?? console.warn(`skipping ${spec.url}, no change`);
+          const copy = Object.assign({}, fallback);
           return expandSpecResult(copy, fallbackFolder);
         }
 
--- a/src/lib/specs-crawler.js
+++ b/src/lib/specs-crawler.js
@@ -29,6 +29,7 @@ const {
     createFolderIfNeeded
 } = require('./util');
 
+const {version: reffyVersion} = require('../../package.json');
 
 /**
  * Return the spec if crawl succeeded or crawl result from given fallback list
@@ -78,13 +79,16 @@ async function crawlSpec(spec, crawlOptions) {
         path.dirname(crawlOptions.fallback) : '';
 
     if (spec.error) {
-        return specOrFallback(spec, fallbackFolder, crawlOptions.fallbackData);
+        return specOrFallback(spec, fallbackFolder, crawlOptions.fallbackData?.results);
     }
 
     try {
-        const fallback = crawlOptions.fallbackData?.find(s => s.url === spec.url);
-        const etag = fallback?.crawlCacheInfo?.etag;
-        const lastModified = fallback?.crawlCacheInfo?.lastModified;
+        let fallback;
+        let cacheInfo = {};
+        if (crawlOptions.fallbackData?.crawler === `reffy-${reffyVersion}`) {
+          fallback = crawlOptions.fallbackData?.results?.find(s => s.url === spec.url);
+          cacheInfo = Object.assign({}, fallback?.crawlCacheInfo);
+        }
         const result = await processSpecification(
             spec.crawled,
             (spec, modules) => {
@@ -101,7 +105,7 @@ async function crawlSpec(spec, crawlOptions) {
             [spec, crawlOptions.modules],
             { quiet: crawlOptions.quiet,
               forceLocalFetch: crawlOptions.forceLocalFetch,
-              etag, lastModified}
+              ...cacheInfo}
         );
         if (result.status === "notmodified" && fallback) {
           crawlOptions.quiet ?? console.warn(`skipping ${spec.url}, no change`);
@@ -195,7 +199,7 @@ async function crawlSpec(spec, crawlOptions) {
         spec.error = err.toString() + (err.stack ? ' ' + err.stack : '');
     }
 
-    return specOrFallback(spec, fallbackFolder, crawlOptions.fallbackData);
+    return specOrFallback(spec, fallbackFolder, crawlOptions.fallbackData?.results);
 }
 
 
@@ -363,7 +367,7 @@ async function crawlList(speclist, crawlOptions) {
     // Load fallback data if necessary
     if (crawlOptions.fallback) {
         try {
-            crawlOptions.fallbackData = JSON.parse(await fs.promises.readFile(crawlOptions.fallback)).results;
+            crawlOptions.fallbackData = JSON.parse(await fs.promises.readFile(crawlOptions.fallback));
         } catch (e) {
             throw new Error(`Could not parse fallback data file ${crawlOptions.fallback}`);
         }
@@ -481,12 +485,14 @@ async function saveResults(data, settings) {
 
     // Save all results to an index.json file
     const indexFilename = path.join(settings.output, 'index.json');
+
     const contents = {
         type: 'crawl',
         title: 'Reffy crawl',
         date: (new Date()).toJSON(),
         options: settings,
         stats: {},
+        crawler: `reffy-${reffyVersion}`,
         results: data
     };
     contents.options.modules = contents.options.modules.map(mod => mod.property);
--- a/src/lib/specs-crawler.js
+++ b/src/lib/specs-crawler.js
@@ -103,7 +103,7 @@ async function crawlSpec(spec, crawlOptions) {
               forceLocalFetch: crawlOptions.forceLocalFetch,
               etag, lastModified}
         );
-        if (result.error === "reuseexistingdata" && fallback) {
+        if (result.status === "notmodified" && fallback) {
           crawlOptions.quiet ?? console.warn(`skipping ${spec.url}, no change`);
           const copy = Object.assign({}, fallback);
           return expandSpecResult(copy, fallbackFolder);
--- a/src/lib/specs-crawler.js
+++ b/src/lib/specs-crawler.js
@@ -178,7 +178,9 @@ async function crawlSpec(spec, crawlOptions) {
 
         // Copy results back into initial spec object
         spec.crawled = result.crawled;
-        spec.crawlCacheInfo = result.crawlCacheInfo;
+        if (result.crawlCacheInfo) {
+          spec.crawlCacheInfo = result.crawlCacheInfo;
+        }
         crawlOptions.modules.forEach(mod => {
             if (result[mod.property]) {
                 spec[mod.property] = result[mod.property];
--- a/src/lib/specs-crawler.js
+++ b/src/lib/specs-crawler.js
@@ -52,6 +52,7 @@ async function specOrFallback(spec, fallbackFolder, fallbackData) {
         if (fallback) {
             const copy = Object.assign({}, fallback);
             const result = await expandSpecResult(copy, fallbackFolder);
+            result.error = spec.error;
             return result;
         }
     }
--- a/src/lib/specs-crawler.js
+++ b/src/lib/specs-crawler.js
@@ -403,7 +403,6 @@ async function crawlList(speclist, crawlOptions) {
         const spec = specAndPromise.spec;
         const logCounter = ('' + (idx + 1)).padStart(nbStr.length, ' ') + '/' + nbStr;
         crawlOptions.quiet ?? console.warn(`${logCounter} - ${spec.url} - crawling`);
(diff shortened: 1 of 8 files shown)


dontcallmedom:
this depends on https://github.com/tidoust/fetch-filecache-for-crawling/pull/6

from my exploration, this can reduce a no-change crawl to ~1min30; but there are server-side configuration issues (on www.w3.org and rfc-editor.org, probably on csswg.org although I haven't diagnosed them yet) that makes the actual impact as is less big at the moment.

dontcallmedom:
@tidoust if you get a chance to review the current direction, this would be very much appreciated

src/lib/specs-crawler.js
```diff
@@ -82,6 +84,9 @@ async function crawlSpec(spec, crawlOptions) {
     }
 
     try {
+        const fallback = (crawlOptions.fallbackData || []).find(s => s.url === spec.url);
```
tidoust, line 87:
```suggestion
        const fallback = crawlOptions.fallbackData?.find(s => s.url === spec.url);
```

src/lib/specs-crawler.js
```diff
@@ -392,6 +403,7 @@ async function crawlList(speclist, crawlOptions) {
         const spec = specAndPromise.spec;
         const logCounter = ('' + (idx + 1)).padStart(nbStr.length, ' ') + '/' + nbStr;
         crawlOptions.quiet ?? console.warn(`${logCounter} - ${spec.url} - crawling`);
+        crawlOptions.logCounter = logCounter;
```
tidoust, line 406: I'm surprised that this works given that several specs are crawled in parallel. You'll need to make a shallow copy of `crawlOptions` or use another mechanism to pass that parameter.

src/lib/util.js
```diff
@@ -497,14 +529,27 @@ async function processSpecification(spec, processFunction, args, options) {
...
+            }
+            const responseHeaders = result.headers();
+            if (responseHeaders.etag) {
+              cacheInfo = {etag: responseHeaders.etag};
+            } else if (responseHeaders['last-modified']) {
+              cacheInfo = {lastModified: responseHeaders['last-modified']};
             }
+          } catch (err) {
+            if (reuseExistingData) {
+              throw new ReuseExistingData(`${spec.url} hasn't changed`);
```
tidoust, line 550: I'm not super fan of using exceptions for handling a case that is not going to be exceptional at all. I would prefer if this was handled as part of the regular code flow.

I haven't looked into how easy it may be to take another approach though. There is probably no other way to do it at the network interception level, but once we catch the case here, could we not throw and rather somehow report that the result of processing the spec is that the spec does not need to be processed?

src/lib/specs-crawler.js
dontcallmedom: it doesn't work, and was removed from my local copy but I failed to update the committed code :)

src/lib/util.js
dontcallmedom: I agree using exceptions here is bad, fixed

dontcallmedom:
Thanks for the review, I've fixed the bugs you've mentioned and improved a bit the cache semantics to avoid relying on etag; the CSS server is still problematic (which is unfortunate given their many many specs), but I'm down to 3min30 of processing despite that.

dontcallmedom:
Added another option to skip processing that accounts for mis-behaving servers; it now runs in 1min30 with still a few servers/spec not providing cache info (fxtf, some random rfcs)

src/lib/util.js
```diff
@@ -497,13 +531,30 @@ async function processSpecification(spec, processFunction, args, options) {
...
             await page.setContent(spec.html, loadOptions);
         }
         else {
-            const result = await page.goto(spec.url, loadOptions);
+            let result;
+            try {
+              result = await page.goto(spec.url, loadOptions);
+            } catch (err) {
+              if (reuseExistingData) {
+                return {error: "reuseexistingdata"};
```
tidoust, line 544: That's more editorial than anything else, but following on my previous comment not to use an exception, I would prefer if that wasn't reported as an "error" but rather as a status response (there was no error, the result of processing is the equivalent of a 304 in HTTP).

```suggestion
                return {status: "notmodified"};
```

(We could also return a `{status: "ok"}` in cases where processing takes place and `{status: "error"}` when processing fails, but let's focus on the problem at hand)

src/lib/specs-crawler.js
```diff
@@ -97,8 +100,14 @@ async function crawlSpec(spec, crawlOptions) {
             },
             [spec, crawlOptions.modules],
             { quiet: crawlOptions.quiet,
-              forceLocalFetch: crawlOptions.forceLocalFetch }
+              forceLocalFetch: crawlOptions.forceLocalFetch,
+              etag, lastModified}
         );
+        if (result.error === "reuseexistingdata" && fallback) {
```
tidoust, line 106: See comment on using `status` instead of `error`:
```suggestion
        if (result.status === "notmodified" && fallback) {
```

tidoust (changes requested):
A couple of nits.

This looks good to me otherwise. I note that one implicit hypothesis here is that the index page of a multipage spec is always going to be updated when one of the pages gets updated. I think that's the case because the index page contains the last modified date. I'm not sure that is always the case though. For instance, is the index page always updated when multiple updates get made to pages in the same day?

More importantly, as-is, this caching mechanism (not surprisingly?) creates a cache invalidation problem: if we fix or update one of the extraction modules in `browserlib` or even other parts of Reffy's code that could affect extracts, and continue to run Reffy with the fallback parameter, the update will only be reflected in extracts when specs get modified.

That problem already exists without this caching mechanism but is limited to specs that cannot be crawled (which we hope is going to be a temporary condition).

The obvious solution is to force a full crawl when that happens. However, from a Webref perspective:

1. We don't really know which version of Reffy was used to crawl data, so we cannot easily run Reffy without the fallback parameter when a new version comes out.
2. We probably want to keep the fallback parameter in any case, so as to reuse previous crawl results in case of errors and not introduce temporary updates in the extracts.

I'm thinking that a relatively simple solution would be to inject the version of Reffy that was used to crawl the data in `index.json` (which is useful info to have in any case), and to only enable this caching mechanism when the current version matches the one in the fallback crawl result.

dontcallmedom:
> I'm thinking that a relatively simple solution would be to inject the version of Reffy that was used to crawl the data in `index.json` (which is useful info to have in any case), and to only enable this caching mechanism when the current version matches the one in the fallback crawl result.

I had been thinking along the same line, although I had imagined we would do only a full crawl when reffy was bumped in minor or major versions (while having a dispatchable workflow to do a full crawl on demand as well)

Do you want this in this PR, or can this be done separately?

tidoust:
> Do you want this in this PR, or can this be done separately?

I would prefer if we had it in this PR so as not to start a loop in Webref that cannot be fixed until we get the feature. But if that's the next thing you do, I'm fine doing it separately as well

(I still wonder about multipage specs though)

dontcallmedom:
I've added support for checking reffy version to use fallback data; this could made more subtle (i.e. not necessarily require an exact match), but it's probably enough as starting point.

>  I note that one implicit hypothesis here is that the index page of a multipage spec is always going to be updated when one of the pages gets updated. I think that's the case because the index page contains the last modified date. I'm not sure that is always the case though. For instance, is the index page always updated when multiple updates get made to pages in the same day?

Looking at multi-page specs we have today:
* HTML, CSS2.2 and ES are generated from a single source file, so the title page is guaranteed to be updated
* CSS 2.1 and SVG 1.1 are TR-only (so would necessary have their title page updated if they ever get updated)
* SVG2 is problematic - its title page DOES NOT get updated when subpages are updated

In practice, since we would still get a fresh data update on each reffy upgrade, and given that SVG2 gets updated very rarely, I'm not too concerned by that limitation; we could check when fetching subpages whether their last-modified is (significantly) more recent than the title page and not save the cache info in that situation if we really care.

src/lib/specs-crawler.js
```diff
@@ -78,13 +79,16 @@ async function crawlSpec(spec, crawlOptions) {
...
 
     try {
-        const fallback = crawlOptions.fallbackData?.find(s => s.url === spec.url);
-        const etag = fallback?.crawlCacheInfo?.etag;
-        const lastModified = fallback?.crawlCacheInfo?.lastModified;
+        let fallback;
+        let cacheInfo = {};
+        if (crawlOptions.fallbackData?.crawler === `reffy-${reffyVersion}`) {
+          fallback = crawlOptions.fallbackData?.results?.find(s => s.url === spec.url);
+          cacheInfo = Object.assign({}, fallback?.crawlCacheInfo);
```
tidoust, line 90: I would still use the fallback data when versions don't match, so as not to end up with dangling extract files in case of processing errors. I would just not set `cacheInfo` to force the processing.

dontcallmedom: hmm... but if the format of the extract files have changed for a reason or another, I'm not sure re-using them is a good idea

tidoust:
> HTML, CSS2.2 and ES are generated from a single source file, so the title page is guaranteed to be updated

But will that generation always trigger an actual git update? What is going to change in the title page? The date is per day in particular, so would the title page get git-updated the second time a spec receives an update on the same day? That situation is probably not that uncommon for the HTML spec.

src/lib/specs-crawler.js
tidoust: But that situation is meant to be temporary. If we don't reuse them, the dangling extract files will be removed from the package, triggering a pre-release PR to drop them (and a minor bump).

dontcallmedom: I'm not sure I understand; say we change the JSON schema in idlparsed extracts, the said extracts with the previous schema are no longer consumable, so we shouldn't pretend we have the relevant data for the said spec.

dontcallmedom:
> But will that generation always trigger an actual git update? What is going to change in the title page? The date is per day in particular, so would the title page get git-updated the second time a spec receives an update on the same day? That situation is probably not that uncommon for the HTML spec.

For HTML, the title page includes a link with the relevant commit (for the snapshot) - so if that's our primary concern, I think we're safe

src/lib/specs-crawler.js
tidoust: Well, we've always been pretending that we have the relevant data. We keep dangling extract files in the repo for the time being in the hope that they will still be accurate next time processing succeeds. They appear in the curated branch (in your example, not reusing the extract files won't change anything, the potentially outdated idlparsed extracts won't be dropped from the repo and will continue to appear in the curated branch in particular) and they are released in the npm packages as well.

The fallback mechanism gives us a way to keep the links from `index.json` to these dangling extract files, so that we effectively no longer have any dangling extract files in the repo. That's great, I don't like dangling extract files because I don't know what they mean.

The release mechanism now only considers extract files that are linked from `index.json` (not strictly needed today, but will be needed as soon as we need to filter non-browser specs).

When the version of Reffy changes, if we don't reuse the extract files, the pre-release PR will say "these files have been removed", which is not fantastic because the right message should rather be "we have not up-to-date extract files for these specs, and don't know whether they are still accurate". Also, we won't be able to release a new version of a package until we're back to a crawl without error situation.

If we reuse the extract files, the pre-release PR will not report anything about these files, which is not fantastic either because then we could miss the fact that some of the extract files may no longer have the right structure.

Actually, I think that we need to update the pre-release PR job to include in the diff the list of specs for which we only have fallback data (those that have an `error` property). Even if the structure of the extract files has not changed, there may be good reasons not to release a package if we realize that an extract is too outdated. I'm just worried that we'll often end up with a list of >100 CSS specs. I'll look into it.

In any case, I would prefer to stick to a "no dangling extract files" principle.

dontcallmedom: ok, now the fallback is only skipped for not-modified situations when crawler versions don't match

dontcallmedom:
> Edit: Or are the generated pages not stored in and served from a Git branch?

HTML and CSS2 are not served from a git branch afaict; ES is, but it looks (from a cursory check) that the title page is updated for the date when one of the subpages is https://github.com/tc39/ecma262/commit/65905e90e4ff67a4aa9440bedf2bbbde4c69ec78#diff-bf7934c87fd1a409fa27452a6461d0ca2916decce82025154e60df5a3e0e8215  - we might lose some same-day updates, but I don't think that warrants a strong concern. This presumably could be addressed with my earlier proposal discussed in the context of SVG2

tidoust (approved):
Good! Go go go!

PR #858, opened 2022-02-04
Check cache validity before launching browser tab

dontcallmedom:
saves memory and time

### DIFF
--- a/src/lib/util.js
+++ b/src/lib/util.js
@@ -21,6 +21,8 @@ const reffyModules = require('../browserlib/reffy.json');
  */
 const maxPathDepth = 20;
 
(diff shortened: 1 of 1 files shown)


dontcallmedom:
This builds on #856 but brings another optimization: this time the primary fetch gets executed before any browser tab is opened and outside the browser networking stack; this reduces the memory consumption and the time needed (this goes down to 55s when no change have to be parsed).

I'm making this a separate PR so as to not block the already long #856, and because I suspect the request headers might need to be improved (at the very with support for `Accept-Encoding`) -  it may be even preferable to use the exact request headers from Puppeteer somehow.

tidoust (approved):

PR #869, opened 2022-02-08
Switch from browser-specs to web-specs

tidoust:
The `web-specs` package contains non browser specs on top of the specs that target browsers in the `browser-specs` package.

Whether a spec targets browsers or not will appear in the crawl results under the `categories` property.

Follows from https://github.com/w3c/browser-specs/issues/436 and https://github.com/w3c/browser-specs/pull/496

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,20 +1,20 @@
 {
   "name": "reffy",
-  "version": "6.2.2",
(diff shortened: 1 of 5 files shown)


dontcallmedom (approved):

dontcallmedom:
Tests need an update

PR #890, opened 2022-02-23
Add support for permission dfn-type

dontcallmedom:
As discussed in https://lists.w3.org/Archives/Public/spec-prod/2022JanMar/0008.html

### DIFF
--- a/src/browserlib/extract-dfns.mjs
+++ b/src/browserlib/extract-dfns.mjs
@@ -34,7 +34,7 @@ function normalize(str) {
 }
 
 // Valid types defined in https://tabatkins.github.io/bikeshed/#dfn-types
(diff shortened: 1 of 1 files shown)


src/browserlib/extract-dfns.mjs
```diff
@@ -68,6 +68,7 @@ function hasValidType(el) {
     'setlike',
     'extended-attribute',
     'event',
+    'permission',
```
tidoust, line 71: Could you also update the comment on line 37 to add "permission" to the list of types not yet defined in Bikeshed's doc?

dontcallmedom: done in 7889fb8

tidoust (approved):

PR #892, opened 2022-02-25
Do not produce CSS/IDL extracts for fork specs

tidoust:
This update makes it possible to integrate "fork" specs in `web-specs` (see https://github.com/w3c/browser-specs/pull/528). Reffy will not produce CSS/IDL extracts for these specs for the time being to avoid running into thorny duplicate issues, see comment in https://github.com/w3c/browser-specs/issues/511#issuecomment-1050958822

### DIFF
--- a/src/lib/specs-crawler.js
+++ b/src/lib/specs-crawler.js
@@ -447,11 +447,20 @@ async function adjustExtractsPerSeries(data, property, settings) {
 
     data.forEach(spec => {
         if (fullLevels.includes(spec)) {
-            // Full level, rename the extract after the series' shortname
-            const pathname = path.resolve(settings.output, spec[property]);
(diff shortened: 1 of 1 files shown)


dontcallmedom (commented):
code LGTM, but an alternative would be to still collect the fragments, but mark them as to be ignored one way or another for downstream processors?

tidoust:
I don't disagree. That's meant as a short term measure as I don't have the bandwidth right now to do something smarter. In Reffy, collecting the fragments means also updating the `idlnames` logic to ignore them. In webref, that means understanding what to do during curation and what to release in `@webref/idl`.

dontcallmedom:
I'll go with this as it has the merit to exist, and to open the way to merge https://github.com/w3c/browser-specs/pull/528; we can revisit iteratively.

PR #970, opened 2022-06-17
Fix to event extraction

dontcallmedom:
bugs identified when compiling the list of gaps in https://github.com/w3c/webref/issues/630#issuecomment-1159028458


tidoust (commented):
Looks good to me with a couple of problems:
- There's a missing call to `replaceWith` before the call to return line 90.
- Indentation, yuk!

I'll push an update to fix these problems.

Issue #971, opened 2022-06-18
CSS terminal values extracted as non-terminals


cdoublev:
For example, `annotation(<feature-value-name>)` is extracted as `"<annotation()>": { "value": "annotation(<feature-value-name>)" }` from [CSS Fonts 4](https://drafts.csswg.org/css-fonts-4/#annotation) but `<annotation()>` does not appear in any value definition.

`annotation(<feature-value-name>)` is marked up with `data-dfn-type="value"` and according to https://github.com/w3c/csswg-drafts/issues/7380#issuecomment-1159125937, it should be treated the same as a keyword:

> While our functions are often uniquely named, they should in fact be treated the same as keywords - potentially context-sensitive.

I think it should be extracted as `"annotation(<feature-value-name>)": { "prose": "Enables display of alternate annotation forms (font specific, OpenType feature: nalt <feature-index>)" }` or just ignored (like keywords).

Note that some non-terminals looks like function, eg. `<calc()>`, and some terminals are annoted between angled brackets, eg. `<number>`. [`<image-set()>`](https://drafts.csswg.org/css-images-4/#funcdef-image-set) is an example of a function definition (with `data-dfn-type="function"`) referenced as a production ([terminal - CSS basic data type - or non-terminal wrapped between `<` and `>`](https://drafts.csswg.org/css-values-4/#component-types)) from another definition, [`<image>`](https://drafts.csswg.org/css-images-4/#typedef-image):

> `<image> = <url> | <image()> | <image-set()> | <cross-fade()> | <element()> | <gradient>`

Note that in its definition, `<image-set()>` is not wrapped between `<` and `>`, which may seem confusing...:

> `image-set() = image-set( <image-set-option># )`

...whereas eg. `<calc()>` is defined with `data-link-type="function"` as:

> `<calc()>  = calc( <calc-sum> )`

tidoust:
Yes, we could look into `data-dfn-type` to create entries that go into `valuespaces`, and e.g. only consider the ones defined with a `data-dfn-type="function"`. That would provide more consistent results with fewer dangling values, even though we would miss a number of entries that are currently defined as `value` whereas it seems to me that they should rather be `function`. Typically, I would expect to see `annotation()` defined as `function`, and referenced through `<annotation()>` from the definition of `font-variant-alternates` so as to be able to create a link between to the definition of `annotation`, as done in some other CSS specs.

[...]

PR #976, opened 2022-06-23
Re-write CSS extraction module based on "data-dfn-type"

tidoust:
This re-writes the CSS extraction module to rely on the `data-dfn-type` attribute, as discussed in #971.

Only production definitions marked with `type` or `function` get extracted.

As an exception to the rule, `<pre>` tags are parsed separately because, while CSS specs consistently use dfns in other forms of production rules, they sometimes mix dfns and references in `<pre>` tags, as in:
https://drafts.csswg.org/css-values-4/#calc-syntax

When definition appears in prose, the paragraph is copied.

I checked the result against webref. This does produce better extracts in the sense of:
- It creates fewer dangling values that don't really exist as such in the spec, or at least that CSS specs don't define as non-terminal values
- It adds values defined in prose, along with a textual representation of the definition. The textual representation is probably not going to be extremely useful but it's at least useful when one reviews the extract. The values themselves will be useful to study "holes" in the grammar.

There are a few things that are not necessarily correct but, as far as I can tell, this is a spec bug and not a bug in the extraction logic. For instance, The [`AnimationEvent` constructor](https://drafts.csswg.org/css-animations/#funcdef-animationevent) is defined as a CSS function in CSS Animations and thus is going to appear in the CSS extract, but that's a spec mistake. I propose to handle these over time through our usual mechanisms.

Note CSS patches in Webref will mostly likely need to be updated as a result of this update.

### DIFF
--- a/src/browserlib/extract-cssdfn.mjs
+++ b/src/browserlib/extract-cssdfn.mjs
@@ -178,143 +178,132 @@ const extractDlDfns = (doc, className, options) =>
 
 /**
  * Extract value spaces (non-terminal values) defined in the specification
+ *
+ * From a definitions data model perspective, non-terminal values are those
+ * defined with a `data-dfn-type` attribute equal to `type` or `function`. They
+ * form (at least in theory) a single namespace across CSS specs.
+ *
+ * Definitions with `data-dfn-type` attribute set to `value` are not extracted
+ * on purpose as they are typically namespaced to another construct (through a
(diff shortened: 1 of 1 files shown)


dontcallmedom (approved):

cdoublev:
FYI, the issue on the `AnimationEvent()` constructor has been [resolved](https://github.com/w3c/csswg-drafts/pull/7454).

PR #990, opened 2022-06-30
Extract interface of event for CSS Animations & CSS Transitions

dontcallmedom:


src/browserlib/extract-events.mjs
```diff
@@ -310,20 +310,37 @@ export default function (spec) {
...
+      // CSS Animation & Transitions document the event in the heading
+      // of the section where the definitions are located
+      let currentEl = container.parentNode;
+      while(currentEl) {
+	if (currentEl.tagName.match(/^H[1-6]$/)) {
+	  break;
+	}
+	currentEl = currentEl.previousElementSibling;
+      }
+      const interfaceEl = currentEl.querySelector("code");
```
tidoust, line 329:
```suggestion
      const interfaceEl = currentEl?.querySelector("code");
```

(We may have to process other specs that use `<dt>` and do not have a section heading at the same level (e.g. because definition table is wrapped in a `<div>` for some reason)

src/browserlib/extract-events.mjs
```diff
@@ -310,20 +310,37 @@ export default function (spec) {
...
+      }
     }
     const ev = events.find(e => isSameEvent(event, e));
     if (!ev) {
+      if (iface) {
+	event.interface = iface;
+      }
       event.bubbles = bubbles;
       events.push(event);
       console.error(`[reffy] No interface hint found for event definition ${event.type} in ${spec.title}`);
```
tidoust, line 341:
```suggestion
      if (!iface) {
        console.error(`[reffy] No interface hint found for event definition ${event.type} in ${spec.title}`);
      }
```

src/browserlib/extract-events.mjs
```diff
@@ -310,20 +310,37 @@ export default function (spec) {
...
     } else {
+      ev.interface = iface;
```
tidoust, line 343:
```suggestion
      if (iface) {
        ev.interface = iface;
      }
```

tidoust (commented):
Looks good with inline suggestions.

(I'll re-process the code once we're done with all updates to use consistent indentation throughout, but just wish we'd be able to fix the issue once and for all...)

PR #1014, opened 2022-07-14
Add 1mn timeout for Respec generation

tidoust:
Respec will never be able to generate a spec that has a syntax error in its Respec configuration. In such cases, Respec will never resolve the "ready" promise.

This update adds a 1mn timeout to make sure that the crawler does not get stuck waiting forever.

(Updated prompted by a [config bug in the PNG spec](https://github.com/w3c/PNG-spec/pull/141) that makes the crawler loop forever)


tidoust:
Taking the liberty to merge so that Webref crawling jobs may resume

PR #1015, opened 2022-07-14
Add post-processor and events post-processing module 

tidoust:
This is a breaking change. It does two main things:
1. It adds a post-processing stage to Reffy. This makes it possible and easy to run modules that may need to mix data extracted by multiple crawling modules, and/or to mix data generated from other specs. The update converts logic that used to be hardcoded in the crawler such as idlparsed generation to post-processing modules.
2. It adds an "events" post-processing module to consolidate event extracts.

More precisely, changes are:
- Add a post-processor library that can run post-processing modules either at the spec level or at the crawl level.
- Convert generateIDLParsed/saveIDLParsed and generateIDLNames/saveIDLNames to post-processing modules "idlparsed" and "idlnames". The former runs at the spec level. The latter runs at the crawl level.
- Create a "csscomplete" post-processing module for post-crawl CSS completing tasks that were previously hardcoded in the crawler.
- Create an "events" post-processing module that integrates the non-patch logic done in webref's PR https://github.com/w3c/webref/pull/650
- Add a `--post` option to Reffy's CLI to specify the post-processing modules to run. None are run by default. This means that "idlparsed" and "idlnames" are no longer generated by default during a crawl. This also means that CSS properties are not completed (e.g. with the IDL attributes they generate) by default either.
- Add a `--crawl` option to Reffy's CLI to make it skip the crawl and rather ingest a crawl result. The option is only really useful when combined with the new `--post` option. It makes it possible to run post-processing modules on an existing crawl result.
- Make the crawler skip the crawl step when the `--crawl` option is used.
- Drop the "generate-idlparsed" and "generate-idlnames" CLI tools. Same result may now be achieved by running the "idlparsed" and "idlnames" post-processing modules through Reffy's CLI.
- Expose the post-processor as an entry in `index.js`.

Bugs fixed along the way:
- The `isLatestLevelThatPasses` function could crash due to a check being done too early.
- The crawler was not setting the right `crawled` property initially.
- The events extraction expected `crawled` to be set on the spec object, but it is typically set "after" the crawler is done running the modules. The window.location URL should be used instead. Tests adjusted accordingly.

Still missing (could be done later on?):
- Post-processor tests.
- Update README to explain the post-processor.
- Smarter workflow logic to order the post-processing modules based on what they depend on. In practice, that's not needed for now because the only theoretical hiccup is between "idlparsed" and "idlnames" and the former runs at the spec level, so the latter always runs after the former.


tidoust:
I ran the crawler locally and confirms that the new version produces the same extracts as the previous ones.

A few additional notes:
1. The `events` post-processing module generates an `events.json` file in the root output folder and not an `index.json` file under an `events` sub-folder. We'll have to adjust the code that will generate the `@webref/events` package accordingly but that should be a no-brainer.
2. A consequence of 1. is that we could actually envision getting back to `events` for the name of the folder that contains the raw events extracts, instead of `spec-events`.
3. Webref's curation code will need to be adjusted to the new version in any case to run the post-processing modules on the patched results.

dontcallmedom:
> 2\. A consequence of 1. is that we could actually envision getting back to `events` for the name of the folder that contains the raw events extracts, instead of `spec-events`.

+1 to that idea

PR #1018, opened 2022-07-18
Fix exported functions crawlSpecs and getInterfaceTreeInfo

tidoust:
The `getTreeInfo` entry point should have been renamed into `getInterfaceTreeInfo`.

The `crawlSpecs` function targeted the `crawlList` crawl function instead of `crawlSpecs`. I'm not clear that was intended and it would be easier for the curation job in webref to directly call `crawlSpecs`. Let's use the major version bump opportunity to fix this. I added a `crawlList` exported function in case someone actually used that entry point.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,6 @@
 module.exports = {
   parseIdl: require("./src/cli/parse-webidl").parse,
   crawlSpecs: require("./src/lib/specs-crawler").crawlSpecs,
-  crawlList: require("./src/lib/specs-crawler").crawlList,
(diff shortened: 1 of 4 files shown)


dontcallmedom (approved):
FWIW `crawlSpecs` pointing to `crawlList` dates back from the origin the package https://github.com/w3c/reffy/pull/454 ; I recall changing the name (because crawlList feels a bit unspecific), but I don't recall why I picked that method over `crawlSpecs`

Issue #1028, opened 2022-07-25
Extract full CSS at-rule preamble and syntax


pyoor:
Currently, the only thing related to CSS at-rules that is extracted are the descriptors associated with a particular at-rule.  The problem with this, is that ignores any syntax related to the pre-amble and body of the at-rule.  For instance, the CSS @counter-style rule, defines the following syntax:

`@counter-style <counter-style-name> { <declaration-list> }`

The `<counter-style-name>` valuespace is recorded, but not in the context of the at-rule syntax.  The `<declaration-list>` is comprised solely of descriptors in `<key>: <value>` pairs as per https://drafts.csswg.org/css-syntax-3/#typedef-declaration-list.  It'd be great if this this information is also extracted, and that the data is stored in a similar fashion to that of [mdn/data](https://github.com/mdn/data/blob/main/css/at-rules.json).  

For example:

```json
{
  "@counter-style": {
    "value": "@counter-style <counter-style-name> { <declaration-list> }",
    "descriptors": {
      "additive-symbols": {
        "for": "@counter-style",
        "value": "[ <integer> && <symbol> ]#"
      },
      "fallback": {
        "for": "@counter-style",
        "value": "<counter-style-name>"
      },
      "negative": {
        "for": "@counter-style",
        "value": "<symbol> <symbol>?"
      },
      "pad": {
        "for": "@counter-style",
        "value": "<integer> && <symbol>"
      },
      "prefix": {
        "for": "@counter-style",
        "value": "<symbol>"
      },
      "range": {
        "for": "@counter-style",
        "value": "[ [ <integer> | infinite ]{2} ]# | auto"
      },
      "speak-as": {
        "for": "@counter-style",
        "value": "auto | bullets | numbers | words | spell-out | <counter-style-name>"
      },
      "suffix": {
        "for": "@counter-style",
        "value": "<symbol>"
      },
      "symbols": {
        "for": "@counter-style",
        "value": "<symbol>+"
      },
      "system": {
        "for": "@counter-style",
        "value": "cyclic | numeric | alphabetic | symbolic | additive | [ fixed <integer>? ] | [ extends <counter-style-name> ]"
      }
    }
  }
}
```

tidoust:
Thanks for the suggestion, @pyoor. I prepared an update to the CSS extraction code in #1033 that should take care of generating the appropriate structure (I just did not index descriptors by name).

[...]

PR #1033, opened 2022-07-29
Extract CSS at-rules, move descriptors under at-rules

tidoust:
This creates an `atrules` property in CSS extracts that contains at-rules syntax as suggested in #1028. Descriptors associated with an at-rule (and all selectors are associated with an at-rule in practice) now appear under the at-rule definition.

The new `atrules` property contains a `value` property when a formal syntax could be extracted from the spec, and a `descriptors` array that lists descriptors associated with the at-rule (if any).

**Breaking change:** The former `descriptors` property that appeared at the root level of the extract no longer exists. The same info can be found under the `atrules` property.

This update also adjusts a couple of tests and adds two new tests on at-rules extraction.

The code can only extract the formal syntax of an at-rule if it is defined in a `<pre class="prod">` block. That is the case in most CSS specs, but some of the definitions are done in `<pre>` tags without any class for now (e.g. `@counter-style` and `@scroll-timeline`). The specs need fixing or the extraction logic needs to also parse mere `<pre>` tags (but that seems fragile).

As opposed to the structure proposed in #1028, the list of descriptors is not indexed by the descriptor's name.

For instance, the `@container` at-rule would lead to:

```json
{
  "@container": {
    "value": "@container [ <container-name> ]? <container-condition> { <stylesheet> }",
    "descriptors": [
      {
        "name": "width",
        "for": "@container",
        "value": "<length>",
        "type": "range"
      },
      {
        "name": "height",
        "for": "@container",
        "value": "<length>",
        "type": "range"
      },
      {
        "name": "inline-size",
        "for": "@container",
        "value": "<length>",
        "type": "range"
      },
      {
        "name": "block-size",
        "for": "@container",
        "value": "<length>",
        "type": "range"
      },
      {
        "name": "aspect-ratio",
        "for": "@container",
        "value": "<ratio>",
        "type": "range"
      },
      {
        "name": "orientation",
        "for": "@container",
        "value": "portrait | landscape",
        "type": "discrete"
      }
    ]
  }
}
```

Creating the pull request as draft as I still need to run additional tests to make sure that the code successfully extracts all known at-rules defined in CSS specs (or that the specs need fixing).

### DIFF
--- a/src/browserlib/extract-cssdfn.mjs
+++ b/src/browserlib/extract-cssdfn.mjs
@@ -13,15 +13,37 @@ import informativeSelector from './informative-selector.mjs';
 export default function () {
   let res = {
     properties: extractTableDfns(document, 'propdef', { unique: true }),
-    descriptors: extractTableDfns(document, 'descdef', { unique: false }),
+    atrules: {},
     valuespaces: extractValueSpaces(document)
   };
+  let descriptors = extractTableDfns(document, 'descdef', { unique: false });
 
   // Try old recipes if we couldn't extract anything
   if ((Object.keys(res.properties).length === 0) &&
-      (Object.keys(res.descriptors).length === 0)) {
+      (Object.keys(descriptors).length === 0)) {
     res.properties = extractDlDfns(document, 'propdef', { unique: true });
-    res.descriptors = extractDlDfns(document, 'descdef', { unique: false });
+    descriptors = extractDlDfns(document, 'descdef', { unique: false });
+  }
+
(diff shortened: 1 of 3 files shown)


tidoust:
Marking the PR as ready for review as tests confirm that the code correctly extracts anything that can be extracted (see https://github.com/w3c/reffy/issues/1028#issuecomment-1204187717).

dontcallmedom (approved):
I didn't expect this would land so fast, amazing!

PR #1042, opened 2022-08-17
Adapt CSS extract saving to new data model using atrules

dontcallmedom:
avoid the problem identified in https://github.com/w3c/webref/pull/698#pullrequestreview-1075707809


tidoust (approved):
Ah, that indeed explains why the crawler manages to extract the at-rule but fails to save it. Thanks!

I note that the `check-missing-dfns.js` script also needs updating (see #1043)

PR #1055, opened 2022-09-01
Restructure links extracts to keep original links

dontcallmedom:
Currently, reffy "canonicalize" urls before recording them in the links extracts.
While that canonicalization is useful to regroup links that points to the same spec, it loses critical information (e.g. for links to multipage specs) and produces information that does not really reflect the reality of the spec, making it hard to produce reliable analysis.

The patch instead records the original link untouched, and annotates it with the spec that can be identified after canonicalization.

This means the structure of links extract will change.


src/lib/specs-crawler.js
```diff
@@ -294,7 +294,7 @@ async function saveSpecResults(spec, settings) {
 async function crawlList(speclist, crawlOptions) {
     // Make a shallow copy of crawl options object since we're going
     // to modify properties in place
-    crawlOptions = Object.assign({}, crawlOptions);
+    crawlOptions = Object.assign({speclist}, crawlOptions);
```
tidoust, line 297: We've never been very clean on separating inputs and outputs. Unless I missed a copy somewhere, we directly store the results of the crawl within the spec objects in `speclist`, at least temporarily while the results get saved to disk. This means `speclist` may also contain a few expanded crawl results (or all of them when `useCrawl` is set or when the user does not specify an output folder), up to >50MB of data.

Here, we're asking Puppeteer to send the list to a separate Chrome process, so Puppeteer will need to serialize/deserialize the list each time. That works but that seems like a waste of processing time and memory.

I think we should rather make a deeper copy here and create a new list of spec objects that only contain core metadata about each spec (and that won't be touched by the crawl). That will still send >500KB of data but that should be fine.

I can live with it, but it could perhaps be argued that the extraction code should remain atomic in the sense that its results should only depend on information about the spec and on the spec's contents (extraction code is not the easiest bit to debug since it runs in a headless Chrome instance, so better keep the logic simple). An alternative would be to split the extraction of links into two parts:
1. The extraction of the raw links in Chrome through `extract-links.mjs` without attempting to link them back to specs.
2. A post-processing module that associates extracted links with specs and runs in the Node.js process. Post-processing is a good place to run logic that needs to look at other specs.

dontcallmedom: That approach makes more sense indeed; I've updated the PR to that effect

src/postprocessing/annotate-links.js
```diff
@@ -0,0 +1,74 @@
...
+    else {
+        return (equivalentUrls ? equivalentUrls : canon);
+    }
+}
+
+const needsSaving = {};
+
+module.exports = {
+  dependsOn: ['links'],
+  input: 'crawl',
```
tidoust, line 41: Ah, the post-processor is a bit clunky here. This post-processing module can run directly once the spec has crawled so `input: 'spec'` would be the right choice in theory (this would also allow to use the default `save` logic that the post-processor contains instead of re-defining it because the post-processor does something else for crawl-wide post-processing modules).

That said, to be able to do that, you would need a way to access the list of specs in the crawl, which could either be done by updating the signature of the run function to `run: async function(spec, specs, options)` or by copying `speclist` into the options as you had done previously (no need to worry about memory copies a priori).

Anyway, no big deal for now. We can revisit that later on.

tidoust (approved):
That looks good to me, thanks for the update!

See inline comments, regarding Strudy in particular coz' this will break it...

tidoust:
Forgot to add: we'll need to update webref's crawl job to run the post-processing module in [`update-ed.yml`](https://github.com/w3c/webref/blob/ff53edeb2bc5c05b467cc61f418f54e55b944901/.github/workflows/update-ed.yml#L29)

PR #1063, opened 2022-09-13
Sort the consolidated list of events

### LINKED ISSUE #1059: Events post-processor: sort events
The list of events created by the events post-processor is more or less ordered by the the order of the specs in browser-specs. Once in a while, the URL or some metadata info about the spec can change that changes that order. This creates artificial diffs for the list of events.

See https://github.com/w3c/webref/pull/722 for an example.

The list of events could rather be ordered by itself, meaning without depending on the order of something that is external to the list.

tidoust:
This would fix #1059.

Note this will trigger a rather dumb release for the `@webref/events` package, but that should happen only once...

### DIFF
--- a/src/postprocessing/events.js
+++ b/src/postprocessing/events.js
@@ -71,6 +71,35 @@ module.exports = {
         cleanTargetInterfaces(event, parsedInterfaces);
         delete event.spec;
         return event;
+      })
+      .sort((event1, event2) => {
+        if (event1.type < event2.type) {
+          return -1;
+        }
+        else if (event1.type > event2.type) {
+          return 1;
+        }
+        else if (event1.interface < event2.interface) {
+          return -1;
+        }
+        else if (event1.interface > event2.interface) {
+          return 1;
+        }
+        else if (!event2.href) {
+          return -1;
+        }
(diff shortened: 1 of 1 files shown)


src/postprocessing/events.js
```diff
@@ -71,6 +71,35 @@ module.exports = {
         cleanTargetInterfaces(event, parsedInterfaces);
         delete event.spec;
         return event;
+      })
+      .sort((event1, event2) => {
+        if (event1.type < event2.type) {
```
dontcallmedom, line 76: this could be simplified as `event1.type.localeCompare(event2.type) || event1.interface.localeCompare(event2.interface)`…

dontcallmedom (approved):
LGTM with a proposed code simplification

src/postprocessing/events.js
tidoust: Conceptually, I tend to avoid `localeCompare` because behavior depends on localization and the host's locale could well be different on a GitHub server and on our computers. Granted, that shouldn't change anything here given the characters used in `type` and `interface`. I'll switch to `localeCompare` but will make the locale explicit.

PR #1075, opened 2022-09-25
Add JSON schemas for extracts export validation function

tidoust:
This creates JSON schemas for all extracts and files generated by Reffy. A new `getSchemaValidationFunction` function is now exported by Reffy to validate data against the schemas. It is typically intended for use in Webref to validate curated data before publication.

Some minor adjustments made along the way to make data more consistent and keep schemas relatively simple:
- The dfns extractor now always returns a `heading` property, defaulting to the top of the page
- The events extractor now gets rid of null properties altogether
- The events post-processor no longer outputs null href.

Existing tests on extractors were updated to also check the schema of the extracted data. More tests could be added to check post-processors, and the crawler as a whole.

Note this update also includes a minor bug fix in the sorting code of the events post-processor so that it may run on not-yet-curated events extracts.

This is based on (and intends to replace the current form of) https://github.com/w3c/webref/pull/731

### DIFF
--- a/schemas/browserlib/extract-dfns.json
+++ b/schemas/browserlib/extract-dfns.json
@@ -1,0 +1,50 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "$id": "https://github.com/w3c/reffy/blob/main/schemas/browserlib/extract-dfns.json",
+
+  "type": "array",
+  "items": {
+    "type": "object",
+    "additionalProperties": false,
+    "required": ["id", "href", "linkingText", "localLinkingText",
+      "type", "for", "access", "informative", "heading", "definedIn"],
+    "properties": {
+      "id": { "$ref": "../common.json#/$defs/id" },
+      "href": { "$ref": "../common.json#/$defs/url" },
+      "linkingText": {
+        "type": "array",
+        "items": { "type": "string" }
+      },
+      "localLinkingText": {
+        "type": "array",
+        "items": { "type": "string" }
+      },
+      "type": { "type": "string" },
+      "for": {
+        "type": "array",
+        "items": { "type": "string" }
+      },
+      "access": {
+        "type": "string",
+        "enum": ["private", "public"]
+      },
+      "informative": {
+        "type": "boolean"
+      },
+      "heading": {
+        "type": "object",
+        "additionalProperties": false,
+        "required": ["href", "title"],
+        "properties": {
+          "id": { "$ref": "../common.json#/$defs/id" },
+          "href": { "$ref": "../common.json#/$defs/url" },
+          "title": { "type": "string" },
+          "number": { "$ref": "../common.json#/$defs/headingNumber" }
+        }
+      },
+      "definedIn": {
+        "type": "string"
+      }
+    }
+  }
+}
--- a/schemas/browserlib/extract-dfns.json
+++ b/schemas/browserlib/extract-dfns.json
@@ -19,7 +19,26 @@
         "type": "array",
         "items": { "type": "string" }
       },
-      "type": { "type": "string" },
+      "type": {
+        "type": "string",
+        "enum": [
+          "property", "descriptor", "value", "type",
+          "at-rule", "function", "selector",
+
+          "namespace", "interface", "constructor", "method", "argument",
+          "attribute", "callback", "dictionary", "dict-member", "enum",
+          "enum-value", "exception", "const", "typedef", "stringifier",
+          "serializer", "iterator", "maplike", "setlike", "extended-attribute",
+          "event", "permission",
+
+          "element", "element-state", "element-attr", "attr-value",
+
+          "scheme", "http-header",
+
+          "grammar", "abstract-op", "dfn"
+        ],
+        "$comment": "Types taken from src/browserlib/extract-dfns.mjs"
+      },
       "for": {
         "type": "array",
         "items": { "type": "string" }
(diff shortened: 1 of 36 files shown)


schemas/common.json
```diff
@@ -0,0 +1,109 @@
...
+  "$schema": "http://json-schema.org/schema#",
+  "$id": "https://github.com/w3c/reffy/blob/main/schemas/common.json",
+
+  "$defs": {
+    "url": {
+      "type": "string",
+      "format": "url"
+    },
+
+    "nullableurl": {
```
dontcallmedom, line 11: seems like this is is no longer used?

schemas/common.json
```diff
@@ -0,0 +1,109 @@
...
+    },
+
+    "cssValue": {
+      "type": "string",
+      "minLength": 1
+    },
+
+    "interface": {
+      "type": "string",
+      "pattern": "^[A-Z]([A-Za-z0-9_])*|console$",
```
dontcallmedom, line 52: this allows any name ending with "console", doesn't it? probably not a big deal though

schemas/browserlib/extract-dfns.json
```diff
@@ -0,0 +1,50 @@
...
+      "href": { "$ref": "../common.json#/$defs/url" },
+      "linkingText": {
+        "type": "array",
+        "items": { "type": "string" }
+      },
+      "localLinkingText": {
+        "type": "array",
+        "items": { "type": "string" }
+      },
+      "type": { "type": "string" },
```
dontcallmedom, line 22: shouldn't we list the dfn types here?

schemas/files/index.json
```diff
@@ -0,0 +1,40 @@
...
+        "errors": {
+          "type": "integer"
+        }
+      }
+    },
+
+    "results": {
+      "type": "array",
+      "items": {
+        "type": "object"
```
dontcallmedom, line 36: ha!

schemas/postprocessing/events.json
```diff
@@ -0,0 +1,52 @@
...
+        "type": "array",
+        "items": {
+          "type": "object",
+          "additionalProperties": false,
+          "required": ["target"],
+          "properties": {
+            "target": { "$ref": "../common.json#/$defs/interface" },
+            "bubbles": { "type": "boolean" },
+            "bubblingPath": {
+              "type": "array",
```
dontcallmedom, line 23: could use `$defs/interfaces` instead

schemas/postprocessing/idlparsed.json
```diff
@@ -0,0 +1,67 @@
...
+        "idlNames": {
+          "type": "object",
+          "propertyNames": { "$ref": "../common.json#/$defs/interface" },
+          "additionalProperties": {
+            "type": "object",
+            "additionalProperties": true,
+            "required": ["fragment", "type"],
+            "properties": {
+              "fragment": { "type": "string" },
+              "type": { "type": "string" }
```
dontcallmedom, line 35: I believe this is same type as the enum defined in idlnames-parsed - maybe worth lifting it to commons?

dontcallmedom (approved):
gosh, another amazing PR :)

a few suggestions, but feel free to merge as is an consider them as input for later :)

src/lib/util.js
```diff
@@ -1005,6 +1009,83 @@ function getInterfaceTreeInfo(iface, interfaces) {
...
+        (schemaName === 'idlnamesparsed') ? path.join('postprocessing', 'idlnames-parsed.json') :
+        (schemaName === 'idlparsed') ? path.join('postprocessing', 'idlparsed.json') :
+        schemaName.startsWith('extract-') ? path.join('browserlib', `${schemaName}.json`) :
+        schemaName.endsWith('.json') ? path.join('postprocessing', schemaName) :
+        path.join('files', 'extracts', `${schemaName}.json`);
+    let schema;
+    try {
+        schema = require(path.join(schemasFolder, schemaFile));
+    }
+    catch (err) {
```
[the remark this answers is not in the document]
tidoust: It threw in the first version I had, but then it seemed a bit rude to throw when the caller requested an unknown schema. Typically, returning `null` allows to avoid a `try... catch` in Webref:
https://github.com/w3c/webref/pull/731/files#diff-128accda0c348bbd52246995174cc7409be8cf4308f3a1389b5d9e47a556cb55R15-R18

schemas/common.json
tidoust: Ah, good point, some parentheses are missing or `^` and `$` need to be repeated. I'll fix that because it also currently allows interfaces that end with clunky characters!

src/lib/util.js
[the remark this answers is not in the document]
dontcallmedom: why it matters - to avoid a situation where you think you're all good because you didn't get a complaint when in fact you made a type in your schema name.

how - I assume we know the patterns and the names that can be applied to it - anything that doesn't match that should throw (but my assumption may very well be wrong :)

src/lib/util.js
```diff
@@ -1026,14 +1026,31 @@ function getInterfaceTreeInfo(iface, interfaces) {
  *   if the requested schema does not exist.
  */
 function getSchemaValidationFunction(schemaName) {
+    // Helper function that selects the right schema file from the given
```
dontcallmedom, line 1029: you may question-mark as much as you want, it actually does improve readability :)

src/lib/util.js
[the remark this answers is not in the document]
tidoust: I was trying not to hardcode the list of names so that we don't have to update that piece of code when we add more extraction tools and post-processing modules. The code can tell which names are correct for extraction modules by looking at `property` in `src/browserlib/reffy.json`. There is no automatic way to tell what files post-processing modules that operate at the crawl level may generate though.

I'll take the liberty to merge as-is. If you think that deserves an update, the floor is yours :)

PR #1110, opened 2022-11-09
Prevent duplicates in dfns extracts

tidoust:
This makes the dfns extraction code skip over a re-defined term (same linking text, same type, same namespace), and report a warning when a duplicate dfn is encountered and skipped.

Warnings are easily glossed over but we don't really have a better mechanism in place right now to report issues found during extractions. Apart from making the whole extraction fail that is, which seems overkill.

Fixes https://github.com/w3c/webref/issues/783


dontcallmedom (approved):
LGTM, although I think the right model would be to use webref curation (which would solve the "nobody looks at warnings" aspect)

tidoust:
I created https://github.com/w3c/webref/issues/789 to track the possibility to do curation on dfns.

PR #1123, opened 2022-11-28
Fix CSS file schema

tidoust:
The file schema should also have been aligned with the new CSS extract structure introduced in v11.0.0


tidoust:
Simple fix, taking the liberty to merge to release a new patch version of Reffy that can be used in Webref.

PR #1125, opened 2022-11-28
Adjust CSS extract schemas and logic

tidoust:
A few incorrect assumptions about CSS extracts that needed to be revisited.


dontcallmedom (approved):
a child selector, who knew!

PR #1129, opened 2022-11-30
CSS extracts: handle function dfns on right hand side of production rules

tidoust:
Via https://github.com/w3c/webref/issues/808#issuecomment-1331836527

Once in a while, function definitions appear in the right hand side of a production rule, as in css-easing-2:

```
<linear-easing-function> = linear(<linear-stop-list>)
```
https://drafts.csswg.org/css-easing-2/#funcdef-linear

The code did not extract the function parameters in such cases, leading to an reported object without a `value` property:

```json
{
  "name": "linear()",
  "type": "function"
}
```

This update fixes that, meaning the final extract will now contain:

```json
{
  "name": "linear()",
  "type": "function",
  "value": "linear(<linear-stop-list>)"
}
```

### DIFF
--- a/src/browserlib/extract-cssdfn.mjs
+++ b/src/browserlib/extract-cssdfn.mjs
@@ -586,6 +586,7 @@ const extractTypedDfn = dfn => {
   const dfnType = dfn.getAttribute('data-dfn-type');
   const dfnFor = dfn.getAttribute('data-dfn-for');
   const parent = dfn.parentNode.cloneNode(true);
+  const fnRegExp = /^([a-zA-Z_][a-zA-Z0-9_\-]+)\([^\)]+\)$/;
 
   // Remove note references as in:
   // https://drafts.csswg.org/css-syntax-3/#the-anb-type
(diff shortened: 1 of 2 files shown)


src/browserlib/extract-cssdfn.mjs
```diff
@@ -607,7 +607,17 @@ const extractTypedDfn = dfn => {
       // Don't attempt to parse pre tags at this stage, they are tricky to
       // split, we'll parse them as text and map them to the right definitions
       // afterwards.
-      res = { name: getDfnName(dfn) };
+      // That said, we may be looking at a function definition on the right hand
+      // side of a production rule, as in the definition of "linear()" in
+      // css-easing-2: https://drafts.csswg.org/css-easing-2/#funcdef-linear
+      // In such a case, we still want to extract the function parameters
+      if (dfn.textContent.trim().match(/^[a-zA-Z_][a-zA-Z0-9_\-]+\([^\)]+\)$/)) {
```
dontcallmedom, line 614:
```suggestion
      const matchFunctionDefinition =  dfn.textContent.trim().match(/^[a-zA-Z_][a-zA-Z0-9_\-]+\([^\)]+\)$/);
      if (matchFunctionDefinition) {
```

src/browserlib/extract-cssdfn.mjs
```diff
@@ -607,7 +607,17 @@ const extractTypedDfn = dfn => {
...
+        const fn = dfn.textContent.trim().match(/^([a-zA-Z_][a-zA-Z0-9_\-]+)\([^\)]+\)$/)[1];
```
dontcallmedom, line 615:
```suggestion
        const fn = matchFunctionDefinition[1];
```
regexp are painful enough to read, so I suggest DRYing them

dontcallmedom (approved):
LGTM with a minor readability suggestion

src/browserlib/extract-cssdfn.mjs
tidoust: Done. FYI, the first and second regexps were different because the second one used a capturing group while the first one did not. The same logic also appeared elsewhere in the code.

PR #1156, opened 2022-12-15
Force interface to `Element` for EPUB elements

tidoust:
I hadn't noticed that elements extraction now also extracts the elements of the Open Container Format from the EPUB specification: https://github.com/w3c/webref/blob/main/ed/elements/epub-33.json

That specification does not specify any API to access the elements. Using the base `Element` class as fallback.

### DIFF
--- a/src/browserlib/extract-elements.mjs
+++ b/src/browserlib/extract-elements.mjs
@@ -216,7 +216,12 @@ export default function (spec) {
       // All elements defined in MathML Core
       // use the MathMLElement interface
       if (shortname === "mathml-core") {
-        elInfo.interface = "MathMLElement" ;
+        elInfo.interface = "MathMLElement";
+      }
+      // All elements of the Open Container Format of EPUB
+      // use the Element interface
+      else if (shortname === "epub-33") {
+        elInfo.interface = "Element";
       }
       else {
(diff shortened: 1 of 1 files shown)


src/browserlib/extract-elements.mjs
```diff
@@ -216,7 +216,12 @@ export default function (spec) {
       // All elements defined in MathML Core
       // use the MathMLElement interface
       if (shortname === "mathml-core") {
-        elInfo.interface = "MathMLElement" ;
+        elInfo.interface = "MathMLElement";
+      }
+      // All elements of the Open Container Format of EPUB
+      // use the Element interface
+      else if (shortname === "epub-33") {
+        elInfo.interface = "Element";
```
dontcallmedom, line 224: shouldn't we use null instead (and adapt the schema / tests accordingly)? if there is no expectation that an interface can be used for these, we shouldn't pretend there is.

tidoust: Possibly, I didn't know how to approach this... The Open Container Format is an XML dialect, so I would expect to get `Element` if I create a DOM out of a file that follows this format. I'm fine either way. I'll adjust the schema instead.

tidoust:
Rather done by relaxing Webref test that expected an `interface` property on the first item of all elements extracts, see https://github.com/w3c/webref/pull/823

PR #1159, opened 2022-12-16
Add crude dfns patching mechanism

tidoust:
This completes dfns extraction rules to ignore IDL terms defined in "exclude" blocks (for Trusted Types), and introduces a post-processing module to patch problematic dfns. The post-processing module only patches two specs for the time being: `orientation-event` which redefines `PermissionState` and `portals`, which redefines `MessageEventSource`.

Goal is to only add dfn patches when the dfns create actual referencing issues, not to solve all dfns issues, as that would be hard to maintain on a day to day basis.

(Webref's crawling job will need to be updated to call the new module for this to affect Webref data)

For additional context, see https://github.com/w3c/webref/issues/306 and https://github.com/tabatkins/bikeshed/pull/2426#issuecomment-1353888578

### DIFF
--- a/src/browserlib/extract-dfns.mjs
+++ b/src/browserlib/extract-dfns.mjs
@@ -260,6 +260,8 @@ export default function (spec, idToHeading = {}) {
       return node;
     })
     .filter(hasValidType)
+    // Exclude IDL terms defined in a block that is flagged as to be excluded
+    .filter(node => !node.closest('.exclude'))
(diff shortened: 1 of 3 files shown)


dontcallmedom (approved):
crude, but a reasonable first step :)

The kind of patching done here feels like it could be made declarative using an approach similar to what we're doing for events, if/when we're looking at a longer term patching solution.

PR #1183, opened 2023-01-18
Extract headings without data-dfn-type that look like dfns

tidoust:
We already consider that a `<dfn>` without a `data-dfn-type` has an implicit `"data-dfn-type"="dfn"` attribute. This update extends that principle to headings too, provided these headings have some other definition-related attribute to make it clear that they want to be considered as definitions as well.

This is typically needed for the definition of "transferable object" in HTML, that has both a `data-export` and `data-lt` attribute but no `data-dfn-type` attribute:
https://html.spec.whatwg.org/multipage/structured-data.html#transferable-objects

Without this update, Reffy would extract a private "transferable objects" term. With this update, it extracts a public "transferable object" term as intended.

In practice, the only other heading across all specs that has a definition-related attribute but no `data-dfn-type` is in the Reporting API: https://w3c.github.io/reporting/#generate-report
(That definition is a bit weird because the spec uses `data-algorithm` where we probably would prefer to see `data-lt`, but I propose to live with it for now).

### DIFF
--- a/src/browserlib/extract-dfns.mjs
+++ b/src/browserlib/extract-dfns.mjs
@@ -607,7 +607,7 @@ function preProcessEcmascript() {
 }
 
 function preProcessHTML() {
-  const headingSelector = ':is(h2,h3,h4,h5,h6)[id]:not(:is([data-dfn-type],[data-export],[data-noexport],[data-lt])) dfn';
+  const headingSelector = ':is(h2,h3,h4,h5,h6)[id]:not(:is([data-dfn-type],[data-dfn-for],[data-export],[data-noexport],[data-lt])) dfn';
 
   // we copy the id on the dfn when it is set on the surrounding heading
   document.querySelectorAll(headingSelector)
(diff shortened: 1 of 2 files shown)


src/browserlib/extract-dfns.mjs
```diff
@@ -216,12 +216,36 @@ function definitionMapper(el, idToHeading, usesDfnDataModel) {
...
+    // have an implicit `"data-dfn-type"="dfn"` attribute, provided they also
+    // have some other definition related attribute (because we only want to
+    // extract headings that want to be seen as definitions)
     'dfn[id]:not([data-lt=""])',
     'h2[id][data-dfn-type]:not([data-lt=""])',
     'h3[id][data-dfn-type]:not([data-lt=""])',
     'h4[id][data-dfn-type]:not([data-lt=""])',
     'h5[id][data-dfn-type]:not([data-lt=""])',
-    'h6[id][data-dfn-type]:not([data-lt=""])'
+    'h6[id][data-dfn-type]:not([data-lt=""])',
```
dontcallmedom, line 228: I think it might be time to replace this long list with `:is(h2,h3,h4,h5,h6)[data-dfn-type]:not(data-lt=""])` (and re-use that pattern for `data-dfn-for` and `data-export`)

tidoust: Ah, I simply didn't know what `:is()` was for :)
I shortened the selectors accordingly. They start to look like regular expressions, yummy!

dontcallmedom (approved):

PR #1189, opened 2023-01-26
Describe type of idl object in index of idlnames

dontcallmedom:
Useful to filter the list of objects when reading data e.g. from the Web


tidoust (changes requested):
The [schema](https://github.com/w3c/reffy/blob/main/schemas/postprocessing/idlnames.json) needs to be adjusted as well.

tidoust (approved):

PR #1206, opened 2023-02-16
CSS extracts: ignore asides in production rules

tidoust:
Bikeshed now produces aside notations in pure production rules too. They need to be dropped before theses rules get parsed as text. A new test was added to check the feature.


tidoust:
Merging without review to hopefully fix data in Webref. @dontcallmedom FYI.

PR #1211, opened 2023-02-21
CSS prose extraction: ignore info panel in `<dd>` case too.

tidoust:
Bikeshed now copies info panels (defined in an `<aside>`) next to dfns. These info panels need to be ignored when extracting prose. They were already in most cases, but not when the info panel appears in a `<dd>` that follows the dfn of the CSS value being extracted.

Tests updated accordingly. Note that `<aside>` elements are sectioning content and as such cannot appear in phrasing content. Bikeshed creates this invalid DOM fragment through JavaScript. Tests now do that as well.

### DIFF
--- a/src/browserlib/extract-cssdfn.mjs
+++ b/src/browserlib/extract-cssdfn.mjs
@@ -666,6 +666,10 @@ const extractTypedDfn = dfn => {
           c.remove();
         }
       });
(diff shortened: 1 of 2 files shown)


tidoust:
Merging to stop creation of noise in Webref data...

PR #1212, opened 2023-02-21
Dfns extraction: hardcode exception for "prefixed syntax"

tidoust:
Fixes https://github.com/w3c/webref/issues/882 pending a more proper dfns curation process as envisioned in https://github.com/w3c/webref/issues/789


tidoust:
@dontcallmedom FYI, merging to get the missing dfn. A proper dfns curation mechanism would certainly be better but that's a topic for another day.

PR #1217, opened 2023-02-23
ECMAScript spec: Fix dfn scoping and abstract op extraction

tidoust:
As raised in https://github.com/w3c/webref/issues/888, abstract operations were only extracted using a long form such as:
  `["AllocateArrayBuffer(constructor, byteLength)"]`

Crawler should not create linking text that does not exist in the spec. In this particular case, the spec *does* create additional linking text through the `aoid` attribute, set to the name of the abstract operation without parens, so it seems fine to use that info, even though the `aoid` attribute is flagged as legacy in the Ecmarkup documentation:
https://tc39.es/ecmarkup/#emu-clause-legacy-attributes

This update adds the parens-less variant to abstract operations when the `aoid` attribute is present. Linking text in the previous example becomes:
  `["AllocateArrayBuffer", "AllocateArrayBuffer(constructor, byteLength)"]`

The code that extracted abstract operation had a couple of issues that this update also fixes:

1. Abstract operations that contain digits such as `ToInt32()` were not extracted. Regular expressions used to extract information have been completed accordingly.

2. The code that tried to scope abstract methods to some class was not specific enough. "Environments Records" are defined as a hierarchy of classes, and methods are defined multiple times, once for each class. Extraction merely took the first occurrence, and reported duplication warnings. It now looks for the most specific class.

(Side note that the update also drops the line that set `ltNodefault`, since it was not used anywhere)

Tests adjusted accordingly.

### DIFF
--- a/src/browserlib/extract-dfns.mjs
+++ b/src/browserlib/extract-dfns.mjs
@@ -292,7 +292,9 @@ function preProcessEcmascript() {
     el.appendChild(dfn);
     // set id
     dfn.setAttribute("id", el.parentNode.getAttribute("id"));
-    dfn.dataset.ltNodefault = true;
+    if (el.parentNode.hasAttribute("aoid")) {
+      dfn.setAttribute("aoid", el.parentNode.getAttribute("aoid"));
+    }
     return dfn;
   };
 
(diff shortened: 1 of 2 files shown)


tidoust:
Diff of the resulting extract is too large to fit in the description but it looks good to me. Changes are:
- The abstract operations à la `ToInt32` are now exported
- Linking text of various abstract operations now starts with the parens-less version
- Abstract operations previously associated with "Environment Records" and "Module Records" are now scoped to more specific classes, e.g. "Function Environment Records" or "Cyclic Module Records". Abstract operations that were dropped as duplicates now appear in the extract.

dontcallmedom (approved):

PR #1269, opened 2023-04-22
Events: Skip asides in "fire an event" phrasing

### LINKED ISSUE #1268: Events extraction: crawler confused by info panels in "fire an event" phrasings
Bikeshed now creates aside panels next to event names. These aside panels are inserted without whitespaces.

The [code that extracts such event definitions in Reffy](https://github.com/w3c/reffy/blob/e1acb3ec0b1323098685f87b3a6c87360bf57719/src/browserlib/extract-events.mjs#L167) uses a `Range` that then gets converted to a string and parsed with a regular expression. This makes Reffy append an `Info` text to the event.

Note the node cannot simply be cloned to remove asides because `createRange` requires nodes to be attached to the document. We can probably just temporarily attach the cloned node to the document to work around that limitation.

The [Compat spec](https://compat.spec.whatwg.org/#ref-for-concept-event-fire) triggers the issue, and currently blocks data curation in Webref, see https://github.com/w3c/webref/commit/962084e9927853e2673626d6b673a51f0031d46c#diff-c979d3426024c90a4acc3188c8f639e75116b459847995d1bbd26c80219be3f3

tidoust:
This completes the "fire an event" extraction logic to ignore annotations that Bikeshed may now add (without whitespaces) next to event names.

Note the update also adds a couple of safeguards to while loops that were not properly constrained, although this has likely not have had any impact on extracted info so far.

Fixes #1268

### DIFF
--- a/src/browserlib/extract-events.mjs
+++ b/src/browserlib/extract-events.mjs
@@ -163,13 +163,7 @@ export default function (spec) {
       // Clone and drop possible annotations to avoid extracting asides.
       // (note the need to temporarily add the cloned node to the document
       // so that ranges can be used)
-      let apos = 0;
-      for (let i = 0; i < a.parentNode.children.length; i++) {
-        if (a.parentNode.children[i] === a) {
-          apos = i;
-          break;
-        }
-      }
+      const apos = [...a.parentNode.children].findIndex(c => c === a);
       const container = a.parentNode.cloneNode(true);
       const aclone = container.children[apos];
 
--- a/src/browserlib/extract-events.mjs
+++ b/src/browserlib/extract-events.mjs
@@ -160,24 +160,46 @@ export default function (spec) {
   [...document.querySelectorAll("a")]
     .filter(a => !a.closest(informativeSelector) && isFiringLink(a))
     .forEach(a => {
-      const container = a.parentNode;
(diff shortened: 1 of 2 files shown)


src/browserlib/extract-events.mjs
```diff
@@ -160,24 +160,46 @@ export default function (spec) {
...
     .forEach(a => {
-      const container = a.parentNode;
-      // There can be multiple "fire an event" links in a container
-      // limiting our text parsing to content in between two such links
-      // (or the end of the container if there is only one such link)
+      // Clone and drop possible annotations to avoid extracting asides.
+      // (note the need to temporarily add the cloned node to the document
+      // so that ranges can be used)
+      let apos = 0;
+      for (let i = 0; i < a.parentNode.children.length; i++) {
```
dontcallmedom, line 167: this could be
```js
const apos = [...a.parentNode.children].findIndex(a);
```

dontcallmedom (approved):

src/browserlib/extract-events.mjs
tidoust: Indeed, will make the change. FWIW, rather `const apos = [...a.parentNode.children].findIndex(c => c === a);`

PR #1351, opened 2023-08-04
Fix Chrome install in job through npm ci foreground option

tidoust:
It seems that Puppeteer loses itself in its local cache (under the user's folder on Windows, in `.cache\puppeteer`) when two versions of Chrome are used as dependencies: first version installs correctly, second one ends up with just a manifest file that actually references the first version. Reffy cannot run as a result.

Reffy currently has the issue because it depends on Puppeteer v21.0.1 directly, and indirectly on v20.9.0 through ReSpec. Using the `--foreground-scripts` option seems to fix the install issue locally, probably because that serializes the Chrome installations. Trying to see whether that fixes the job as well.

Even if it does, we need a proper fix: it should be possible to run `npm ci` without options!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,5 +13,5 @@ jobs:
       with:
         node-version: '18'
         cache: 'npm'
-    - run: npm ci
+    - run: npm ci --foreground-scripts=true
     - run: npm test


tidoust:
Serialization fixes the problem, indeed. This led me to look into Puppeteer's installation scripts where I think I've found the source of the issue. I filed https://github.com/puppeteer/puppeteer/issues/10695.

I propose *not* to merge this pull request because we should not need to use the `--foreground-scripts` option, and to rather wait until the bug gets fixed in Puppeteer. If that takes some time, we may want to pin dependencies to "compatible" versions (which in turn depend on the same version of Puppeteer) to resume jobs in Reffy and Webref.

dontcallmedom:
the required puppeteer fix is included in #1359

PR #1377, opened 2023-09-24
[css extraction] Handle subsidiary at-rules

tidoust:
The `@font-feature-values` at-rule can contain subsidiary at-rules, defined in the spec as at-rules scoped to `@font-feature-values`: https://drafts.csswg.org/css-fonts/#font-feature-values-syntax

Reffy was not expecting at-rules to be scoped and extracted them as root at-rules, with a `for` attribute that did not pass schema validation.

With this update, Reffy no longer lists subsidiary at-rules at main at-rules, and rather lists them as descriptors under the main at-rule they are scoped to.

### DIFF
--- a/src/browserlib/extract-cssdfn.mjs
+++ b/src/browserlib/extract-cssdfn.mjs
@@ -30,7 +30,7 @@ export default function () {
     // Note some selectors are re-defined locally in HTML and Fullscreen. We
     // won't import them.
     atrules: extractDfns({
-      selector: 'dfn[data-dfn-type=at-rule]',
+      selector: 'dfn[data-dfn-type=at-rule]:not([data-dfn-for])',
       extractor: extractTypedDfn,
       duplicates: 'reject',
       warnings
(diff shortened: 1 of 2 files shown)


dontcallmedom (approved):

cdoublev:
Do you plan to update your schema and consider them as namespaced rules rather than descriptors in a future version?

tidoust:
> Do you plan to update your schema and consider them as namespaced rules rather than descriptors in a future version?

The plan was more to force push them into the current schema not to have to introduce a breaking change in the schema ;)

Nested rules have a `"type": "at-rule"` property in the extract to distinguish them from more usual descriptors if needed.

If they are a pain to deal with in the current form, we could perhaps handle them differently and either move them back to the `atrules` level with a `for` property, or list them under an `atrules` property within the parent at-rule object.

cdoublev:
It is not that painful. I consume the rules extracted by Reffy only to watch for new/modified rules. I [pointed out](https://github.com/w3c/csswg-drafts/issues/8156#issuecomment-1736658679) that it might be easier to only define rules with prose.

That said, CSS Syntax does not consider descriptors and namespaced at-rules the same (emphasize added):

 > **Descriptors are similar to properties (and are declared with the same syntax**) but are associated with a particular type of at-rule rather than with elements and boxes in the tree.

https://drafts.csswg.org/css-syntax-3/#css-descriptor

Also note that descriptors for `@media-query` are also not really descriptors:

  > While not technically descriptors, media features can be defined with this block as well, if an "mq" class is put on the block. It then only requires Name, For, and Value.

https://speced.github.io/bikeshed/#descdef-block

Fortunately, I do not think something like `@media (min-width: 1px) { min-width: 1px }` will be valid in the future. Otherwise there would be no way to know what is valid in `()` but not in `{}`, or vice-versa.

PR #1398, opened 2023-10-13
Extract git revision from spec when available

dontcallmedom:
bikeshed exports the sha of the git commit used to generate the spec when possible; that information is valuable to link a particular crawled document to a crawl output. See e.g. https://github.com/dontcallmedom/spec-check-hook/issues/11


tidoust (approved):
Would it be worth raising the need in ReSpec as well, at least when driven by the spec-prod action (I suppose ReSpec itself cannot do much about it unless it gets the info somehow), to try to get more specs with the info?

PR #1415, opened 2023-11-08
Add pub-manifest and audiobooks to patch-dfns

tidoust:
The published versions of these specs do not follow the definitions data model. Reffy thinks it's dealing with old specs as a result, and flag the definitions as exported.

See https://github.com/w3c/webref/issues/1072 for details.


dontcallmedom (approved):
I wonder if the pattern ED vs TR is something that could be and would be worth detecting automatically. I guess if that's only 3 exceptions, probably not worth it, but as we expand to more non-browser specs, there may be more such cases that emerge'

Issue #932, opened 2022-04-28
Support "corrections" model in Recommendations, skip "<del>"


tidoust:
Some specifications, published as Recommendations, incorporate proposed corrections. These corrections use `<del>` and `<ins>` to highlight parts that change between the current and the proposed future version of the spec.

Reffy happily extracts dfns, idl, links, etc. in `<del>` and `<ins>` constructs, meaning that the extracts are a merge between the current and future version of the spec. This can obviously create artificial duplicates. Reffy should rather skip `<del>` constructs to create extracts that represent the future version of the spec.

Examples of specs with proposed corrections:
- Web Audio API: https://webaudio.github.io/web-audio-api/#AudioContext
- WOFF 2.0: https://w3c.github.io/woff/woff2/#p1
- CSS Contain 1: https://drafts.csswg.org/css-contain-1/#contain-property (but note we actually have levels 2 and 3 in browser-specs)

PR #1428, opened 2023-11-22
Ignore links insided  elements

dontcallmedom:
With W3C Recs that embed their diff for candidate or proposed corrections, we can expect more and more specs to have links-that-aren't-meant-to-be links, which this patch allows to ignore.

See https://github.com/w3c/strudy/pull/439#pullrequestreview-1557149608 for a concrete example this would help fix


tidoust (approved):
I note that partly addresses #932. There are probably other areas that we need to check to skip del sections more consistently.

PR #1434, opened 2023-11-28
Make refs extraction more flexible and stable, add tests

tidoust:
There were not tests for the extraction of references! 😱

Adding tests revealed that some cases were not properly covered. Not a big deal when extraction missed a few references. More problematic when extraction starts extracting terms it should not. The latter happens when crawling some of the AOM specs that are about to be added to browser-specs, see: https://github.com/w3c/browser-specs/issues/1088

The extraction now also supports flat lists (used in some AOM specs and in few older or external specs) and is more resilient to situations where "references" appears in headings that are *not* about references per se (e.g., a change log that contains a section entitled "Changes to informative references").

The code is not perfect but manages to extract a few references that were previously missed, such as those in ECMA-402. It also works better with AOM specs.

Further improvements are still possible. Typically, the code could be tied with `create-outline.mjs` to scope sections more correctly.

### DIFF
--- a/src/browserlib/extract-references.mjs
+++ b/src/browserlib/extract-references.mjs
@@ -74,9 +74,8 @@ function getExtractionRules(generator) {
  */
 function nextTag(node, name, until) {
   let nextEl = node.nextElementSibling;
-  while (nextEl && nextEl !== until &&
-      ((name === 'heading' && !nextEl.tagName.match(/^(H\d$|HGROUP)$/)) ||
-       (name !== 'heading' && nextEl.tagName !== name.toUpperCase()))) {
+  const selector = name === "heading" ? "h1,h2,h3,h4,h5,h6,hgroup" : name;
+  while (nextEl && nextEl !== until && !nextEl.matches(selector)) {
     nextEl = nextEl.nextElementSibling;
   }
   if (nextEl === until) {
@@ -101,7 +100,7 @@ function parseReferences(referenceList, options) {
   var defaultRef = [], informativeRef = [];
   options = options || {};
   if (referenceList.tagName === "DL") {
-    [...referenceList.childNodes]
+    [...referenceList.children]
       .filter(child => child.tagName === "DT")
       .forEach(function (dt) {
         var ref = {};
@@ -122,9 +121,15 @@ function parseReferences(referenceList, options) {
       });
   }
   else if (referenceList.tagName === "UL") {
-    [...referenceList.childNodes]
+    [...referenceList.children]
       .filter(child => child.tagName === "LI")
       .forEach(function (li) {
+        // The ECMA-402 spec lists nests another list for more atomic
+        // references with "URLs in your face":
+        // https://tc39.es/ecma402/#normative-references
+        // Let's drop nested lists for now to avoid extracting noise
+        // (TODO: consider smarter code or creating an exception to the rule
+        // for ECMA-402)
         li = li.cloneNode(true);
         [...li.querySelectorAll("ul")].map(el => el.remove());
         var anchor = li.querySelector("a[href]");
--- a/src/browserlib/extract-references.mjs
+++ b/src/browserlib/extract-references.mjs
@@ -68,13 +68,19 @@ function getExtractionRules(generator) {
  * @function
  * @private
  * @param {Node} node The DOM node to use as starting point
- * @param {String} name The sibling name to find
+ * @param {String} name The sibling name to find, "heading" to match any heading
+ * @param {Node} until The optional DOM sibling at which to stop no matter what
  * @return {Node} The next sibling with the given name, null if not found
  */
-function nextTag(node, name) {
+function nextTag(node, name, until) {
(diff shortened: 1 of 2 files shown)


src/browserlib/extract-references.mjs
```diff
@@ -68,14 +68,20 @@ function getExtractionRules(generator) {
...
+ * @param {String} name The sibling name to find, "heading" to match any heading
+ * @param {Node} until The optional DOM sibling at which to stop no matter what
  * @return {Node} The next sibling with the given name, null if not found
  */
-function nextTag(node, name) {
+function nextTag(node, name, until) {
   let nextEl = node.nextElementSibling;
-  while(nextEl && nextEl.tagName !== name.toUpperCase()) {
+  while (nextEl && nextEl !== until &&
+      ((name === 'heading' && !nextEl.tagName.match(/^(H\d$|HGROUP)$/)) ||
```
dontcallmedom, line 78: it may be more readable to having something like:
```js
const selector = name === "heading" ? "h1,h2,h3,h4,h5,h6,hgroup" : name;
``` 
and then in the loop condition 
```js
nextEl.matches(selector)
```

src/browserlib/extract-references.mjs
```diff
@@ -94,23 +100,45 @@ function nextTag(node, name) {
...
-      ref.url = url;
-    }
-    if (options.filterInformative &&
-        desc.textContent.match(/non-normative/i)) {
-      return informativeRef.push(ref);
-    }
-    defaultRef.push(ref);
-  });
+  if (referenceList.tagName === "DL") {
+    [...referenceList.childNodes]
```
dontcallmedom, line 104:
```suggestion
    [...referenceList.children]
```
we only care about elements

src/browserlib/extract-references.mjs
```diff
@@ -94,23 +100,45 @@ function nextTag(node, name) {
...
+        }
+        if (options.filterInformative &&
+            desc.textContent.match(/non-normative/i)) {
+          return informativeRef.push(ref);
+        }
+        defaultRef.push(ref);
+      });
+  }
+  else if (referenceList.tagName === "UL") {
+    [...referenceList.childNodes]
```
dontcallmedom, line 125:
```suggestion
    [...referenceList.children]
```

src/browserlib/extract-references.mjs
```diff
@@ -94,23 +100,45 @@ function nextTag(node, name) {
...
+      .filter(child => child.tagName === "LI")
+      .forEach(function (li) {
+        li = li.cloneNode(true);
+        [...li.querySelectorAll("ul")].map(el => el.remove());
```
dontcallmedom, line 129: a comment as to why this is useful would be useful :)

tests/extract-references.js
```diff
@@ -0,0 +1,278 @@
...
+  </li>
+</ul>`,
+    res: {
+      normative: [{ name: "Unicode Locale Data Markup Language (LDML)", url: "https://unicode.org/reports/tr35/" }],
+      informative: []
+    }
+  },
+
+  {
+    title: "does not use a nested URL as URL of a reference",
```
dontcallmedom, line 183: why is that a feature?

tests/extract-references.js
```diff
@@ -0,0 +1,278 @@
...
+  let browser;
+  let extractRefsCode;
+  const validateSchema = getSchemaValidationFunction('extract-refs');
+
+  before(async () => {
+    const extractRefsBundle = await rollup.rollup({
+      input: path.resolve(__dirname, '../src/browserlib/extract-references.mjs')
+    });
+    const extractRefsOutput = (await extractRefsBundle.generate({
+      name: 'extractIds',
```
dontcallmedom, line 249:
```suggestion
      name: 'extractReferences',
```

tests/extract-references.js
```diff
@@ -0,0 +1,278 @@
...
+    browser = await puppeteer.launch({ headless: true });
+  });
+
+  testRefs.forEach(t => {
+    it(t.title, async () => {
+      const page = await browser.newPage();
+      page.setContent(t.html);
+      await page.addScriptTag({ content: extractRefsCode });
+
+      const extractedRefs = await page.evaluate(async () => extractIds());
```
dontcallmedom, line 263:
```suggestion
      const extractedRefs = await page.evaluate(async () => extractReferences());
```

dontcallmedom (approved):

src/browserlib/extract-references.mjs
tidoust: I expanded on why the code is not smart enough without that for [ECMA-402 references](https://tc39.es/ecma402/#normative-references).

tests/extract-references.js
tidoust: I reformulated a bit. That's mostly due to ECMA-402, to avoid extracting URLs that don't have title and that are not fully what the reference is all about, see previous comment.

PR #1444, opened 2023-12-12
[dfns] Add HTML prose definition when possible

tidoust:
Implements the logic discussed in https://github.com/w3c/respec/issues/4522

Ready for review but discussion is still ongoing, so code may still need to change. For instance, the `data-defines` attribute may end up taking another name. Also happy to change the name of the `prose` property in the dfns extract.

For each term defined in the specification being processed, the code now looks for some element flagged with a `data-defines="#term-id"` attribute. If such an element exists, a `prose` property gets added to the definition in the `dfns` extract with the HTML contents of that element.

The code applies some clean up to the HTML markup it attaches to the `prose` property:
- All asides that authoring tools may add here and there get dropped
- Any element that is not a simple block or inline content element gets dropped
- All attributes are dropped

The clean up logic may need refinement over time once we gain experience with actual definitions. Open questions include:

- Should we be stricter, e.g., only allowing `<p>`, `<br>`, and very common inline elements?
- Should we keep `href` attributes (with an absolute URL) for `<a>` elements?
- Should we keep `title` attributes for `<abbr>` elements?
- Should we keep `class` attributes for `<pre>` elements to help with syntax highlighting?
- Should we keep tables? Images?

There is no good mechanism in Reffy to report potential issues encountered during extraction for the time being. In the meantime, warnings get logged when the code bumps into elements that seem surprising in the context of a term definition.

### DIFF
--- a/schemas/browserlib/extract-dfns.json
+++ b/schemas/browserlib/extract-dfns.json
@@ -64,7 +64,7 @@
       "definedIn": {
         "type": "string"
       },
-      "prose": {
+      "htmlProse": {
         "type": "string",
         "minLength": 1
       }
--- a/schemas/browserlib/extract-dfns.json
+++ b/schemas/browserlib/extract-dfns.json
@@ -63,6 +63,10 @@
       },
       "definedIn": {
         "type": "string"
+      },
+      "prose": {
+        "type": "string",
+        "minLength": 1
       }
     }
   }
--- a/src/browserlib/extract-dfns.mjs
+++ b/src/browserlib/extract-dfns.mjs
@@ -122,6 +122,47 @@ function isNotAlreadyExported(dfn, idx, list) {
   return first === dfn;
 }
 
+// Extract the element's inner HTML content, removing any complex structure,
+// so that the result can be injected elsewhere without creating problems.
+function getHtmlProseDefinition(proseEl) {
+  // Apply modifications to a copy of the element
+  proseEl = proseEl.cloneNode(true);
+
+  // Drop asides that authoring tools add here and there
+  let el;
+  const asideSelector = [
+    'aside', '.mdn-anno', '.wpt-tests-block', '.annotation',
+    '[id^=dfn-panel-]'
+  ].join(',');
+  while (el = proseEl.querySelector(asideSelector)) {
+    el.remove();
+  }
+
+  // Keep simple grouping content and text-level semantics elements
+  const keepSelector = [
+    'blockquote', 'dd', 'div', 'dl', 'dt', 'figcaption', 'figure', 'hr', 'li',
+    'ol', 'p', 'pre', 'ul',
+    'a', 'abbr', 'b', 'bdi', 'bdo', 'br', 'cite', 'code', 'data', 'dfn', 'em',
+    'i', 'kbd', 'mark', 'q', 'rp', 'rt', 'ruby', 's', 'samp', 'small', 'span',
+    'strong', 'sub', 'sup', 'time', 'u', 'var', 'wbr'
+  ].join(',');
+  while (el = proseEl.querySelector(`:not(${keepSelector})`)) {
+    // The content is more complex than anticipated. It may be worth checking
+    // the definition to assess whether the extraction logic needs to become
+    // smarter. For lack of a better reporting mechanism for now, let's record
+    // a warning.
+    console.warn('[reffy]', `Unexpected element "${el.nodeName}" found in textual definition of "${proseEl.getAttribute('data-defines')}"`);
+    el.remove();
+  }
+
+  // Drop all attributes
+  [...proseEl.querySelectorAll('*')].forEach(el => {
+    el.getAttributeNames().forEach(attr => el.removeAttribute(attr));
+  });
+
+  return proseEl.innerHTML.trim();
+}
+
 function definitionMapper(el, idToHeading, usesDfnDataModel) {
   let definedIn = 'prose';
   const enclosingEl = el.closest('dt,pre,table,h1,h2,h3,h4,h5,h6,.note,.example') || el;
@@ -157,7 +198,7 @@ function definitionMapper(el, idToHeading, usesDfnDataModel) {
   url.hash = '#' + encodeURIComponent(el.getAttribute('id'));
   const href = url.toString();
 
-  return {
+  const dfn = {
     // ID is the id attribute
     // (ID may not be unique in a multi-page spec)
     id: el.getAttribute('id'),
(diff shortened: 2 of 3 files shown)


src/browserlib/extract-dfns.mjs
```diff
@@ -122,6 +122,47 @@ function isNotAlreadyExported(dfn, idx, list) {
...
+    // The content is more complex than anticipated. It may be worth checking
+    // the definition to assess whether the extraction logic needs to become
+    // smarter. For lack of a better reporting mechanism for now, let's record
+    // a warning.
+    console.warn('[reffy]', `Unexpected element "${el.nodeName}" found in textual definition of "${proseEl.getAttribute('data-defines')}"`);
+    el.remove();
+  }
+
+  // Drop all attributes
+  [...proseEl.querySelectorAll('*')].forEach(el => {
```
dontcallmedom, line 159:
```suggestion
proseEl.querySelectorAll('*').forEach(el => {
```
`forEach` is supported by `NodeList`

src/browserlib/extract-dfns.mjs
```diff
@@ -122,6 +122,47 @@ function isNotAlreadyExported(dfn, idx, list) {
...
+    el.getAttributeNames().forEach(attr => el.removeAttribute(attr));
```
dontcallmedom, line 160: per your open question, it feels like keeping `href` with absolute URLs would make sense to me

src/browserlib/extract-dfns.mjs
```diff
@@ -211,6 +252,17 @@ function definitionMapper(el, idToHeading, usesDfnDataModel) {
     // indicates that definition appears in the main body of the specification)
     definedIn
   };
+
+  // Extract a prose definition in HTML for the term, if available
+  const proseEl = document.querySelector(`[data-defines="#${dfn.id}"]`);
```
dontcallmedom, line 257: is there any reason to use `#` in data-defines? unless we expect to allow for other selectors, having the the id would seem sufficient?

schemas/browserlib/extract-dfns.json
```diff
@@ -63,6 +63,10 @@
       },
       "definedIn": {
         "type": "string"
+      },
+      "prose": {
```
dontcallmedom, line 67: should we name this htmlProse or something of the sort to make it clear this is markup, not plain text?

src/browserlib/extract-dfns.mjs
dontcallmedom: maybe only theoretical, but I suspect keeping `lang` and `dir` would be important; they could be added when the need arises, but I'm also not entirely sure yet how we would notice the need. Maybe via consumers unable to use the content?

src/browserlib/extract-dfns.mjs
```diff
@@ -122,6 +122,47 @@ function isNotAlreadyExported(dfn, idx, list) {
...
+    'aside', '.mdn-anno', '.wpt-tests-block', '.annotation',
+    '[id^=dfn-panel-]'
+  ].join(',');
+  while (el = proseEl.querySelector(asideSelector)) {
+    el.remove();
+  }
+
+  // Keep simple grouping content and text-level semantics elements
+  const keepSelector = [
+    'blockquote', 'dd', 'div', 'dl', 'dt', 'figcaption', 'figure', 'hr', 'li',
```
dontcallmedom, line 143: while having a `<dl>` inside the definition block might work, I don't think it would work as a container of a definition - maybe we should have distinct selectors for validating the content vs validating the container? the same would probably apply with `<ol>`, `<ul>`, and `<table>` if it was added.

src/browserlib/extract-dfns.mjs
dontcallmedom: separately, I wonder if we couldn't handle the "simple" case of `<dt>`/`<dd>` automatically based on a simpler convention (e.g. if `dfn.closest('dt') && dfn.closest('.dfn-desc')` then assumes the `<dd>` is the description of the `<dfn>` in the `<dt>`)? This would match was was being discussed in  https://github.com/w3c/respec/issues/4522 , e.g.
```html
<dl class='dfn-desc'>
  <dt><dfn id='foo'>Foo</dfn></dt>
  <dd>A Foo is a non-sensical word used as a placeholder</dd>
```

(later, we could even have `.dfn-desc` automatically identify the `<dt>` as equivalent to a `<dfn>` - but that probably requires more changes in the authoring tools)

src/browserlib/extract-dfns.mjs
tidoust: Yes, but `map` isn't, which always bugs me when I update code that used `forEach` and for which `map` becomes a better approach. Anyway, I'll update...

src/browserlib/extract-dfns.mjs
tidoust: > is there any reason to use `#` in data-defines? unless we expect to allow for other selectors, having the id would seem sufficient?

No reason. I started without it, but the current proposal has a `#`...

tidoust: > This would match was was being discussed in [w3c/respec#4522](https://github.com/w3c/respec/issues/4522)

Yes, I expect authoring tools will support that convention... and produce the right `data-defines` attribute in the generated spec so that extraction in Reffy can remain without heuristics!

src/browserlib/extract-dfns.mjs
tidoust: Could you clarify what "container" means here?

I would define it as the element with the `data-defines` attribute. And the code does not check the container at all in practice. It can be anything, the code will strip it away and only report its contents in the `prose` (or `htmlProse`) property.

dontcallmedom: you're right, I managed to confuse myself on this

src/browserlib/extract-dfns.mjs
tidoust: Actually, for consistency with the rest of the code in that file, I would prefer to keep the `[...x]` construct. The code uses that in a number of places because it needs to `filter()` the list before it runs `forEach()`.

src/browserlib/extract-dfns.mjs
dontcallmedom: I think it's fine to proceed with the `#` syntax given the reasoning behind it - this would be reasonably easy to simplify later on if needed

src/browserlib/extract-dfns.mjs
tidoust: `dir`, `lang` and `href` (with an absolute URL) are now preserved.
Code also keeps `title` because that seemed potentially useful too for rendering of tooltips and expansion of abbreviations.

dontcallmedom (approved):

PR #1452, opened 2023-12-21
Remove HTML comments from extracting htmlProse for definitions

dontcallmedom:
Avoid bloat, esp. since ReSpec use comments expansively

### DIFF
--- a/src/browserlib/extract-dfns.mjs
+++ b/src/browserlib/extract-dfns.mjs
@@ -138,6 +138,13 @@ function getHtmlProseDefinition(proseEl) {
     el.remove();
   }
 
+  // Remove comments
+  const commentsIterator = document.createNodeIterator(proseEl, NodeFilter.SHOW_COMMENT);
+  let c;
+  while ((c = commentsIterator.nextNode())) {
+    c.remove();
+  }
+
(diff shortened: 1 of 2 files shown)


tidoust (approved):
Truth be told, I thought `createNodeIterator` was a new method. ;)
I'm not so far from the truth, it wasn't in [DOM Level 2 Core](https://www.w3.org/TR/2000/REC-DOM-Level-2-Core-20001113/core.html)!

src/browserlib/extract-dfns.mjs
```diff
@@ -138,6 +138,13 @@ function getHtmlProseDefinition(proseEl) {
     el.remove();
   }
 
+  // Remove comments
+  const commentsIterator = document.createNodeIterator(proseEl, NodeFilter.SHOW_COMMENT);
+  let c;
+  while ((c = commentsIterator.nextNode())) {
+    c.remove();
```
tidoust, line 145: Grumpy @tidoust does not like single character variables.

```suggestion
  let comment;
  while ((comment = commentsIterator.nextNode())) {
    comment.remove();
```

tidoust (changes requested):
That said, it would be good to write a short test about that, I'll add one.

PR #1468, opened 2024-01-25
CSS extracts: Link entries back to spec anchor

tidoust:
Requested in https://github.com/w3c/webref/issues/642

It was possible to link entries in CSS extracts back to entries in the dfns extract with a bit of logic but:
1. This requires having both CSS extracts and dfns extracts at hand, and the dfns extracts typically aren't included in the `@webref/idl` package.
2. This is somewhat error-prone.

This updates the CSS definitions extraction logic to also extract the absolute URL (with fragment) of the definition whenever possible. A slight change of approach was needed as previous extraction logic handled "multi-definitions" (e.g., a table definition that defines multiple properties at once) through a text-based approach, and the code now also needs to track the IDs.

Tests already contained IDs, so this update merely refreshes the expected results.

### DIFF
--- a/schemas/browserlib/extract-cssdfn.json
+++ b/schemas/browserlib/extract-cssdfn.json
@@ -14,6 +14,7 @@
         "required": ["name"],
         "properties": {
           "name": { "$ref": "../common.json#/$defs/cssPropertyName" },
+          "href": { "$ref": "../common.json#/$defs/url" },
           "value": { "$ref": "../common.json#/$defs/cssValue" },
           "newValues": { "$ref": "../common.json#/$defs/cssValue" },
           "values": { "$ref": "../common.json#/$defs/cssValues" },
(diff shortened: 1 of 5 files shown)


tidoust:
This seems to work surprisingly well in practice, in that all entries seem to end up with an `href` property, except when a spec somehow extends a base definition, in other words:

1. Definitions that have a `newValues` key typically don't have an `href` property.
2. At-rules that neither have a `prose` nor a `value` key usually don't have an `href` property.
3. The [`contain`](https://drafts.csswg.org/css-contain-3/#contain-property) property in the css-contain-3 delta spec does not have an `href` property, because the property is defined in css-contain-2. Plus it wouldn't be fun without an exceptional case.

Important note: This adds a property to most entries in CSS extracts, which will invalidate the CSS patches in Webref. We'll have to re-generate them. Not too hard... but painful :/

dontcallmedom (approved):

PR #1495, opened 2024-02-21
Add links back to spec in elements extracts

tidoust:
The elements extracts were the last ones that did not have a link back to the spec. With this update, extraction now also adds an `href` property with a link back to the right fragment in the spec.

This works in all known cases, allowing us to turn that into a package guarantee for `@webref/elements`.


dontcallmedom (approved):
hyperlinks ftw!

PR #1499, opened 2024-02-23
Use util.inspect() to report crawl errors

tidoust:
When a spec cannot be crawled, the `error` property is set to something that looked like a detailed error with a stack trace, but that actually skipped over the chain of causes that led to the final `throw` clause. In the end, when a fetch failed, the fetched URL appeared in the message but not the reason why the fetch failed.

Using `util.inspect()` gives us a more complete error message. That's targeted at debugging, and message may not be stable from one version of Node.js to another, but understanding precisely what went wrong is exactly what we need in such cases!

Practically speaking, I'm hoping that this will help me investigate the (hopefully transient) crawl fetch errors currently reported in Webref for FX and Houdini task force specs.

### DIFF
--- a/src/lib/specs-crawler.js
+++ b/src/lib/specs-crawler.js
@@ -13,6 +13,7 @@
 const fs = require('fs');
 const path = require('path');
 const specs = require('web-specs');
+const inspect = require('util').inspect;
 const cssDfnParser = require('./css-grammar-parser');
 const postProcessor = require('./post-processor');
 const {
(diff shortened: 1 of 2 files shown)


dontcallmedom (approved):

tidoust:
> Practically speaking, I'm hoping that this will help me investigate the (hopefully transient) crawl fetch errors currently reported in Webref for FX and Houdini task force specs.

Of course, the errors magically disappeared as soon as I released a new version of Reffy with this update ;)

PR #1515, opened 2024-03-16
Extract HTML attributes for each element

wbamberg:
I don't know if this is viable really but thought it was worth asking. From a conversation in the MDN Discord I thought we could try to add attributes to the extracts for HTML elements.

This PR thinks a `<dd>` item under "Content attributes:" is an attribute if:

- its first child is a `<code>` element
- whose first child is an `<a>` element
- whose href has a fragment beginning `attr`

This is intended to be quite strict, as I thought it's better to exclude legit attributes than to include nonsense.

Anyway, with this test I find the following items excluded (apart from "Global attributes", which they all have):

*************

[`<link>`](https://html.spec.whatwg.org/multipage/semantics.html#the-link-element)
- "Also, the title attribute has special semantics on this element:  Title of the link;  CSS style sheet set name\n     "

[`<style>`](https://html.spec.whatwg.org/multipage/semantics.html#the-style-element)
- "Also, the title attribute has special semantics on this element:  CSS style sheet set name\n     "

[`<body>`](https://html.spec.whatwg.org/multipage/sections.html#the-body-element)
- "onafterprint",
- "onbeforeprint",
- "onbeforeunload",
- "onhashchange",
- "onlanguagechange",
- "onmessage",
- "onmessageerror",
- "onoffline",
- "ononline",
- "onpageswap",
- "onpagehide",
- "onpagereveal",
- "onpageshow",
- "onpopstate",
- "onrejectionhandled",
- "onstorage",
- "onunhandledrejection",
- "onunload"

[`<li>`](https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element)
- "If the element is not a child of an ul or menu element: value —  Ordinal value of the list item\n     "

[`<a>`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element)
- "ping —  URLs to ping\n     ",

[`<dfn>`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-dfn-element)
- "Also, the title attribute has special semantics on this element:  Full term or expansion of abbreviation\n     "

[`<abbr>`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-abbr-element)
- "Also, the title attribute has special semantics on this element:  Full term or expansion of abbreviation\n     "

[`<bdi>`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-bdi-element)
- "Also, the dir global attribute has special semantics on this element."

[`<bdo>`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-bdo-element)
- "Also, the dir global attribute has special semantics on this element."

[`<embed>`](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-embed-element)
- "Any other attribute that has no namespace (see prose)."

[`<area>`](https://html.spec.whatwg.org/multipage/image-maps.html#the-area-element)
- "ping —  URLs to ping\n     ",

[`<input>`](https://html.spec.whatwg.org/multipage/input.html#the-input-element)
- "Also, the title attribute has special semantics on this element:  Description of pattern (when used with pattern attribute)\n     "

[`<fencedframe>`](https://wicg.github.io/fenced-frame/#elementdef-fencedframe)
- "allow — Permissions policy to be applied to the fencedframe's contents\n    "

**************

All the "Also, the XYZ attribute" seem right to exclude: they're not adding a new attribute but qualifying an existing one. The event handler attributes on `<body>` seem right to exclude, unless there's some reason the spec lists these only for `<body>`.

That leaves us with a few that are a bit trickier:

- [`<a>`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) and [`<area>`](https://html.spec.whatwg.org/multipage/image-maps.html#the-area-element) list `ping`, but the fragment ID does not start with `"attr"`, it's just `"ping"`. I'm not sure if there's a good reason for this or if it is fixable in the spec.
- [`<fencedframe>`](https://wicg.github.io/fenced-frame/#elementdef-fencedframe) lists `allow` but again its fragment does not start with `"attr"`, it's `"element-attrdef-fencedframe-allow"`.
- [`<li>`](https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element) does not list the attribute first - perhaps it could be rewritten like "`value` - If the element is not a child of an ul or menu element, represents the ordinal value of the list item"?
- [`<embed>`](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-embed-element) has "Any other attribute that has no namespace (see prose)." I didn't really understand this, even after referring to the prose. Perhaps it would be OK to omit this? The MDN page for `<embed>` makes no mention of any extra attributes, either.

### DIFF
--- a/src/browserlib/extract-elements.mjs
+++ b/src/browserlib/extract-elements.mjs
@@ -60,7 +60,7 @@ export default function (spec) {
             // 'Contexts in which this element can be used': 'contexts',
             // 'Content model': 'contents',
             // 'Tag omission in text/html': 'omission',
-            // 'Content attributes': 'attributes',
+            'Content attributes': 'attributes',
             // 'Accessibility considerations': 'accessibility',
             'DOM interface': 'interface'
           })[getText(dt).replace(/:$/, '')];
@@ -119,6 +119,25 @@ export default function (spec) {
             }
             else {
               throw new Error('Could not link element to interface, missing dd for ' + res.name);
+            }
+          } else if (prop === "attributes") {
+            res[prop] = [];
+            let dd = dt.nextElementSibling;
+            while (dd && dd.nodeName !== "DT") {
+              if (dd.nodeName === "DD") {
+                // To count as an attribute, we look for:
+                // - the first child is a CODE element,
+                // - and its first child is an A element
+                // - and its href contains a fragment starting with "attr"
+                if (
+                  dd.firstChild?.nodeName === "CODE" &&
+                  dd.firstChild?.firstChild?.nodeName === "A" &&
+                  new URL(dd.firstChild?.firstChild).hash.startsWith("#attr")
+                ) {
+                  res[prop].push(getText(dd.firstChild?.firstChild));
+                }
+              }
+              dd = dd.nextElementSibling;
             }
           }
           else if (prop) {


tidoust:
Thanks for the code and exploration, @wbamberg!

Intuitively, I would have started with a dfn-based approach to extract the list of content attributes: extract all the `<dfn>` that have a `data-dfn-type="element-attr"`, and look at the `data-dfn-for` attribute to associate the content attribute with the element(s) it is defined for. The data already exists in the dfns extracts but could be added to the elements extracts as well.

I think this approach would capture all the attributes that you manage to extract with your code, although as usual it's going to be interesting to compare the exact results of both approaches to reveal problems worth fixing in the specs ;)

The dfn-based approach would also help capture additional content attributes, some of them you mentioned in your "trickier" list:
- The content attributes of [`<portal>`](https://wicg.github.io/portals/#elementdef-portal), whose definition does not use an element definition table at all.
- The `ping` attribute of [`<a>`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) and [`<area>`](https://html.spec.whatwg.org/multipage/image-maps.html#the-area-element)
- The `allow` attribute of [`<fencedframe>`](https://wicg.github.io/fenced-frame/#elementdef-fencedframe)
- The `value` attribute of [`<li>`](https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element)

This wouldn't be perfect though. For example:
- This would miss the content attributes of the [`<model>`](https://immersive-web.github.io/model-element/#the-model-element) element because they are correctly defined as `element-attr` but don't have a `data-dfn-for` attribute. That should be fixed in the spec.
- This would also miss the content attributes of [MathML Core](https://w3c.github.io/mathml-core/) elements for the same reason.
- This would not work for SVG-related specs that don't follow the usual dfn contract (SVG, SVG Animations, Filter Effects).

> The event handler attributes on <body> seem right to exclude, unless there's some reason the spec lists these only for <body>.

As far as I can tell, the HTML spec lists them for `<body>` because they are specific to `<body>` (all other event handler content attributes are common to all elements, see [Event handlers section in HTML](https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers-on-elements,-document-objects,-and-window-objects)), so I would argue they ought to be included. Interestingly, the HTML spec has one definition for these attributes that is both for the event handlers **and** the content attributes. The `<dfn>` defines an event handler in practice (`data-dfn-type="attribute"), and not a content attribute. The dfn-based approach won't work for them...

In the end, to create reliable data, I think we'll have to combine the approaches (as already done for the elements extraction itself) and improve the dfn markup in a couple of specs.

I'm not quite sure what to do with global attributes. In HTML, they come with a `data-dfn-for="html-global"` which does not target an actual element. It would make sense to include them in the elements extracts somehow, perhaps with a flag that explains that they are not specific to the element?

wbamberg:
Thank you @tidoust for such a helpful response. 

> 
> > The event handler attributes on `<body>` seem right to exclude, unless there's some reason the spec lists these only for `<body>`.
> 
> As far as I can tell, the HTML spec lists them for `<body>` because they are specific to `<body>` (all other event handler content attributes are common to all elements, see [Event handlers section in HTML](https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers-on-elements,-document-objects,-and-window-objects)), so I would argue they ought to be included.

Fair enough, and I see that MDN already lists these things as `<body>` attributes: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/body#onafterprint.

> Interestingly, the HTML spec has one definition for these attributes that is both for the event handlers **and** the content attributes. The `<dfn>` defines an event handler in practice (`data-dfn-type="attribute"`), and not a content attribute. The dfn-based approach won't work for them...
> 
> In the end, to create reliable data, I think we'll have to combine the approaches (as already done for the elements extraction itself) and improve the dfn markup in a couple of specs.
> 

By "combine the approaches" do you mean consider a thing an attribute if it meets _either_ the test in this PR _or_ your dfn-based approach?

I'm happy to have a go at that in this PR if you like.

> I'm not quite sure what to do with global attributes. In HTML, they come with a `data-dfn-for="html-global"` which does not target an actual element. It would make sense to include them in the elements extracts somehow, perhaps with a flag that explains that they are not specific to the element?

So a follow-up for me was: as a consumer of this data, I would prefer to distinguish between HTML, MathML, and SVG, rather than just have "elements". For instance, MDN makes this distinction. Do you think it would be desirable to have webref/html as well as webref/elements? And if so, it feels to me like including all the globals in each element would be unnecessary, and we could instead have a separate `globalAttributes` property, and it can just be understood by consumers that by definition all HTML elements include the global attributes.

tidoust:
> By "combine the approaches" do you mean consider a thing an attribute if it meets either the test in this PR or your dfn-based approach?

Yes. I would start with the dfn-based approach because content attributes ought to follow this convention, and fallback to parsing the table structure only when that fails (or to get the `onxxx` attributes for `<body>`). Note we may also handle exceptions through a "patch", as [already done to capture obsolete elements in HTML](https://github.com/w3c/webref/blob/main/ed/elementspatches/html.json.patch).

> I'm happy to have a go at that in this PR if you like.

Go go go :)

> Do you think it would be desirable to have webref/html as well as webref/elements?

You mean: split the package into multiple packages? I would prefer to avoid that because we manually review the packages before publication (data is curated through the patches I mentioned and we sometimes detect things that look weird at that step). Splitting the data wouldn't create more review work, but would introduce additional manual steps to release the packages. That said, the data in webref/elements is already per spec. Or are you looking at a specific grouping that the current data does not capture? (e.g. `<model>` is in its own spec but is an HTML element in practice).

It seems fine to break the structure of elements extracts if there's no good place to put the global attributes and/or additional info if needed.

tidoust:
As I bumped into it recently, I note a discussion in the issue tracker of the Pulsar text editor (which leverages Webref packages already) that mentions the need to have an easy-to-use list of elements with attributes, for autocomplete purpose: https://github.com/pulsar-edit/pulsar/issues/393#issuecomment-1445229044

(As said above, the data already exists in the dfns extracts of Webref, but then that data is not part of an npm package, so harder to integrate, and then there are a few exceptions that need to be accounted for).

dontcallmedom:
re pulsar, their [autocomplete list](https://github.com/pulsar-edit/pulsar/blob/master/packages/autocomplete-html/update/curated-attributes.json) provides us with a comparison point (for completeness verification) and a possible data model.

PR #1518, opened 2024-03-20
Differentiate automatic links from others in link extractors

dontcallmedom:
Should help with strudy link analysis; maybe also be useful in identifying specs that could readily switch to automatic links


tidoust (approved):
This introduces easy to spot but breaking changes to the structure of links extracts.
Do we know whether there are projects out there, other than ours, that make use of them?

dontcallmedom:
I'm not aware of any

PR #1520, opened 2024-03-21
Fix schema for links extract... and add tests

tidoust:
Grmpf. The schema update recently introduced was invalid. That seemed a good opportunity to add a few tests to make sure that does not happen again ;)


dontcallmedom (approved):
good job :)

Issue #1582, opened 2024-06-06
Re-write crawler's main loop using proper queues


tidoust:
Via @dontcallmedom in https://github.com/w3c/reffy/pull/1581#discussion_r1628097955:

> I wonder how much the interleaving buys us vs managing a proper per-"origin" queue of requests that would let us process requests in parallel without pauses as long as there is no lock on the relevant origins.
>
> I think e.g. https://github.com/w3c/cg-monitor/blob/5ceea24a1eaa7c1736d38909bcbb47df4ae297a3/lib/caching-queued-fetch.js#L68 accomplishes this for the CG monitor

PR #1587, opened 2024-06-06
Re-write throttling logic with queues

tidoust:
(Leaving as draft as this would need to be updated if the code for throttled-queue changes following review of https://github.com/w3c/browser-specs/pull/1356)

This re-writes the convoluted logic introduced in previous commit (and the also convoluted logic that pre-existed) as suggested in #1582, taking inspiration from the [queue in the CG monitor](https://github.com/w3c/cg-monitor/blob/5ceea24a1eaa7c1736d38909bcbb47df4ae297a3/lib/caching-queued-fetch.js#L68).

On top of serializing requests sent to a given origin and sleeping a bit in between these requests, the new queue controls parallelization, making sure that the code does not load too many specs at once (I stuck to 4 in parallel as before).

The net result is essentially the same as that achieved with the previous commit: specs are processed as if origins had been interleaved. But the new code is more straightforward, less verbose and more readable.

Also, the code for the throttled queue is the exact same one as the code in browser-specs (see https://github.com/w3c/browser-specs/pull/1356). Code is duplicated here out of simplicity and laziness to create and maintain a shared dependency.

### DIFF
--- a/src/lib/specs-crawler.js
+++ b/src/lib/specs-crawler.js
@@ -345,7 +345,16 @@ async function crawlList(speclist, crawlOptions) {
         return result;
     }
 
-    const crawlQueue = new ThrottledQueue(4);
+    const crawlQueue = new ThrottledQueue({
+        maxParallel: 4,
+        sleepInterval: origin => {
+            switch (origin) {
+            case 'https://csswg.org': return 2000;
+            case 'https://www.w3.org': return 1000;
+            default: return 100;
+            }
+        }
+    });
     const results = await Promise.all(list.map((spec, idx) => {
         const versionToCrawl = crawlOptions.publishedVersion ?
             (spec.release ? spec.release : spec.nightly) :
--- a/src/lib/specs-crawler.js
+++ b/src/lib/specs-crawler.js
@@ -16,6 +16,7 @@ const specs = require('web-specs');
 const inspect = require('util').inspect;
 const cssDfnParser = require('./css-grammar-parser');
 const postProcessor = require('./post-processor');
+const ThrottledQueue = require('./throttled-queue');
 const {
     completeWithAlternativeUrls,
     expandBrowserModules,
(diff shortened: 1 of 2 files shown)


src/lib/throttled-queue.js
```diff
@@ -0,0 +1,120 @@
...
+ */
+module.exports = class ThrottledQueue {
+  originQueue = {};
+  maxParallel = 4;
+  ongoing = 0;
+  pending = [];
+
+  constructor(maxParallel) {
+    if (maxParallel >= 0) {
+      this.maxPar = maxParallel;
```
dontcallmedom, line 54:
```suggestion
      this.maxParallel = maxParallel;
```

tidoust:
That seems to work well, with the caveat that crawl takes much longer due to the serialization per origin and the 2s pause in-between specs that target the same origin. The thing is there are >300 nightly specs with a `github.io` URL. That means crawl will take ~12minutes at best, closer to 15mn in practice. That's fine in theory.

We could split the different `github.io` origins (I know, I argued against that in browser-specs ;)). We'll still have ~200 specs with a `w3c.github.io` URL, and ~85 specs with a `wicg.github.io` URL, but that's better than grouping them together.

(I'll prepare a side PR to avoid requests on SVG and CSS resources, that works fine as well, and seems useful!)

dontcallmedom:
my personal sense is that we should only use the 2 seconds pacing for the CSS server, and a much smaller one for other "origins"

tidoust:
> my personal sense is that we should only use the 2 seconds pacing for the CSS server, and a much smaller one for other "origins"

Done in the latest commit. I kept a 1 second sleep time for www.w3.org because of the 180 requests/minute limit.

Note that getOrigin now also returns the "https://" prefix, to be more consistent with what the `origin` property returns.

The update does not fundamentally change the duration of a full crawl: loading the specs is what takes the most amount of time. It should speed up the 204 case in incremental crawls though.

src/lib/specs-crawler.js
```diff
@@ -345,7 +345,16 @@ async function crawlList(speclist, crawlOptions) {
         return result;
     }
 
-    const crawlQueue = new ThrottledQueue(4);
+    const crawlQueue = new ThrottledQueue({
+        maxParallel: 4,
+        sleepInterval: origin => {
```
dontcallmedom, line 350: isn't this missing the integration of `sleepInterval` in `ThrottledQueue`?

tidoust: Ah, yes, better with all updates :) I force pushed things.

dontcallmedom (approved):

PR #1588, opened 2024-06-07
Skip requests on SVG and CSS resources

tidoust:
SVG resources were loaded because a couple of specs have a weird PNG fallback mechanism. The Accelerometer spec still had an issue, but that's now fixed, and crawl now works well without having to load SVG resources.

CSS resources were loaded mostly out of fear that not doing so would break something. I skipped them and ran a crawl locally. I could not spot anything problematic.

(I'll report this to browser-specs as well)

### DIFF
--- a/src/lib/mock-server.js
+++ b/src/lib/mock-server.js
@@ -119,18 +119,6 @@ mockAgent
 
 mockAgent
   .get("https://www.w3.org")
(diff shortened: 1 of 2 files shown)


dontcallmedom (approved):

dontcallmedom:
(the tests needs an update though)

PR #1614, opened 2024-07-08
Extract normative algorithms defined in specs

tidoust:
This adds a browserlib module that creates extracts filled with information about algorithms defined in the spec. The extracts are rather raw: they just capture the tree structure of algorithms and copy the HTML for each step. This makes the resulting extracts less directly useful but also makes it possible to run all sorts of analyses on them (see related PR in Strudy: https://github.com/w3c/strudy/pull/645 and the [current results of running the analysis](https://github.com/tidoust/parallel-promise/issues/2)).

This would add about 50MB of additional data to a Webref crawl result. That's significant, roughly equivalent to the IDs extracts, which are the heaviest for now.

The structure of the extracts is very likely going to change substantively as we learn from experience!

There are plenty of things that could be improved. The code contains TODOs for main ones.

An algorithm extract is essentially an object with the following keys:
- `name`: The name of the algorithm, when one exists
- `href`: The URL with fragment to reach the algorithm, when one exists
- `html`: Some introductory prose for the algorithm. That prose may well contain actual algorithmic operations, e.g.: "When invoked, run the following
 steps in parallel". href/src attributes in the HTML have absolute URLs.
- `steps`: Atomic algorithm steps.

Each step is essentially an object that follows the same structure as an algorithm, except that it does not have a `name` and `href` keys, and may also have the following keys:
- `operation`: Gives the name of the main operation performed by the step, when one was identified. So far, that's only for "switch".
- `case`: Used in switch steps to identify the switch condition that triggers the step.
- `ignored`: Ordered lists found at the step level that do no look like algorithm steps. Or maybe they are? The lists should get reviewed: they usually describe inputs/outputs or conditions, but they may signal parts where the extraction logic needs to be improved. The lists are reported as text prose.
- `additional`: Each step should contain one and only one algorithm. When other algorithms are found at the same level, they get reported in that property. That usually either signals that the spec could be improved because if fails to use different list items for different steps, and/or that the extraction logic needs to be smarter.

### DIFF
--- a/src/browserlib/extract-algorithms.mjs
+++ b/src/browserlib/extract-algorithms.mjs
@@ -407,7 +407,7 @@ function findIntroParagraph(algo) {
     // https://w3c.github.io/webappsec-csp/#abstract-opdef-parse-a-serialized-csp
     // TODO: improve!
     paragraph = algo.root;
-    while (paragraph && paragraph.nodeName !== 'P') {
+    while (paragraph && (paragraph.nodeName !== 'P' || paragraph.matches(informativeSelector))) {
       paragraph = paragraph.previousElementSibling;
     }
   }
--- a/src/browserlib/extract-algorithms.mjs
+++ b/src/browserlib/extract-algorithms.mjs
@@ -389,10 +389,29 @@ function getDefinedNameIn(el) {
  * there's one.
  */
 function findIntroParagraph(algo) {
-  let paragraph = algo.root;
-  while (paragraph && paragraph.nodeName !== 'P') {
-    paragraph = paragraph.previousElementSibling;
+  let paragraph;
+  let container = algo.root.closest('.algorithm');
+  while (container) {
+    const dfn = container.querySelector('dfn');
+    if (dfn) {
+      paragraph = dfn.closest('p,div');
+      break;
+    }
+    container = container.parentElement.closest('.algorithm');
   }
+
+  if (!paragraph) {
+    // Consider that the introductory paragraph is the previous paragraph.
+    // That's not going to be 100% correct. For example, we will incorrectly
+    // capture an intermediary paragraph as in:
+    // https://w3c.github.io/webappsec-csp/#abstract-opdef-parse-a-serialized-csp
+    // TODO: improve!
+    paragraph = algo.root;
+    while (paragraph && paragraph.nodeName !== 'P') {
+      paragraph = paragraph.previousElementSibling;
+    }
+  }
+
   return paragraph;
 }
 
@@ -413,6 +432,10 @@ function getAlgorithmInfo(algo, context) {
     if (container && !context?.nested) {
       if (container.getAttribute('data-algorithm')) {
         info.name = normalize(container.getAttribute('data-algorithm'));
+        if (container.getAttribute('data-algorithm-for')) {
+          info.name = normalize(container.getAttribute('data-algorithm-for')) +
+            '/' + info.name;
+        }
         if (container.id) {
           // Use the container ID as anchor
           info.href = getAbsoluteUrl(container);
--- a/src/browserlib/extract-algorithms.mjs
+++ b/src/browserlib/extract-algorithms.mjs
@@ -670,5 +670,5 @@ export default function (spec, idToHeading = {}) {
     // remove duplicate from multi-steps algorithms also detected by the
     // short form algorithms finder (relying on them being later in the
     // list)
-    .filter((algo, idx, arr) => arr.findIndex(al => !algo.html || al.html === algo.html) === idx);
+    .filter((algo, idx, arr) => !algo.html || algo.steps.length > 0 || arr.findIndex(al => al.html === algo.html) === idx);
 }
--- a/src/browserlib/extract-algorithms.mjs
+++ b/src/browserlib/extract-algorithms.mjs
@@ -1,0 +1,663 @@
+/**
+ * Extract normative algorithms defined in specs.
+ *
+ * An algorithm extract is essentially an object with the following keys:
+ * - `name`: The name of the algorithm, when one exists
+ * - `href`: The URL with fragment to reach the algorithm, when one exists
+ * - `html`: Some introductory prose for the algorithm. That prose may well
+ * contain actual algorithmic operations, e.g.: "When invoked, run the following
+ * steps in parallel". href/src attributes in the HTML have absolute URLs.
+ * - `steps`: Atomic algorithm steps.
+ * - `parallel`: A flag set to true when the steps are to run "in parallel".
+ *
+ * Each step is essentially an object that follows the same structure as an
+ * algorithm, except that it does not have a `name` and `href` keys, and may
+ * also have the following keys:
+ * - `operation`: Gives the name of the main operation performed by the step,
+ * for example "switch", "let", "set", "if", "return", "resolve", "reject",
+ * "queue a task", "fire an event", etc.
+ * - `case`: Used in switch steps to identify the switch condition that
+ * triggers the step.
+ * - `ignored`: Ordered lists found at the step level that do no look like
+ * algorithm steps. Or maybe they are? The lists should get reviewed: they
+ * usually describe inputs/outputs or conditions, but they may signal parts
+ * where the extraction logic needs to be improved. The lists are reported as
+ * text prose.
+ * - `additional`: Each step should contain one and only one algorithm. When
+ * other algorithms are found at the same level, they get reported in that
+ * property. That usually either signals that the spec could be improved
+ * because if fails to use different list items for different steps, and/or
+ * that the extraction logic needs to be smarter.
+ *
+ * TODO: flag step operation when understood (queue a task, fire an event,
+ *  run in parallel, etc.) to ease analysis.
+ *  (the property is only set for identified "switch" constructs for now)
+ * TODO: handle "read requests"
+ *  https://fetch.spec.whatwg.org/#incrementally-read-loop
+ *  https://w3c.github.io/webcodecs/#imagedecoder-fetch-stream-data-loop
+ * TODO: handle "fetch" process request/response algorithms
+ *  https://wicg.github.io/background-fetch/#complete-a-record
+ *  https://wicg.github.io/nav-speculation/prefetch.html#create-navigation-params-by-fetching
+ * TODO: support a switch without a ".switch" class
+ *  https://w3c.github.io/webcodecs/#dom-videoframe-videoframe
+ *  https://w3c.github.io/web-nfc/#dfn-map-text-to-ndef
+ * TODO: support a switch that is not phrased as a switch
+ *  https://w3c.github.io/clipboard-apis/#to-os-specific-well-known-format
+ * TODO: support a switch where cases don't have <dd>
+ *  https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-errata-20220621.html#sctn-minpinlength-extension
+ * TODO: don't get confused by conditions that look like steps
+ *  (code reports them as "ignored", that's a good start, ignore them fully!)
+ *  https://w3c.github.io/webcodecs/#imagedecoder-decode-complete-frame
+ *  https://w3c.github.io/presentation-api/#dom-presentationrequest-start
+ *  https://w3c.github.io/clipboard-apis/#dom-clipboard-read
+ * TODO: don't get confused by informative "algorithms"
+ *  (noting informative sections are not flagged as such in Bikeshed)
+ *  https://drafts.csswg.org/css-view-transitions-2/#lifecycle
+ * TODO: convert branching operations to substeps when needed ("if")
+ *  https://drafts.css-houdini.org/css-layout-api-1/#construct-a-fragment-result
+ *  https://w3c.github.io/webappsec-credential-management/#dom-passwordcredential-store-slot
+ *  https://dom.spec.whatwg.org/#concept-create-element
+ * TODO: don't get confused by intermediary notes that jeopardize steps lists
+ *  (but then, the specs need fixing!)
+ *  https://w3c.github.io/secure-payment-confirmation/#sctn-steps-to-check-if-a-payment-can-be-made
+ *  https://w3c.github.io/ServiceWorker/#on-fetch-request-algorithm
+ *  https://wicg.github.io/turtledove/#dom-navigator-createauctionnonce
+ * TODO: convert inline operations to substeps when needed
+ * TODO: filter out CSS algorithms that are not JS algorithms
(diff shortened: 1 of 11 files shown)


src/browserlib/extract-algorithms.mjs
```diff
@@ -0,0 +1,663 @@
...
+  }
+
+  return rationale?.toString();
+}
+
+
+/**
+ * Return true if element has the given parent as ancestor
+ */
+function hasParent(el, parent) {
```
dontcallmedom, line 586: this sounds like a rewrite of [`Node.contains()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains)

dontcallmedom (commented):
wow!

results from a first review (hoping that's useful); if you could not force-commit moving forward, that'd be helpful :)

I hope this will be coming with tests given all the funky archaeology you must have done on this!

src/browserlib/extract-algorithms.mjs
```diff
@@ -0,0 +1,663 @@
...
+ * perform.
+ *
+ * The following list of verbs is used to assess whether a set of steps "looks
+ * like" a set of algorithm steps, so as to avoid extracting lists that are not
+ * algorithms.
+ *
+ * The list is completed with a few branching operations that are not verbs:
+ * "for", "if", "while".
+ *
+ * Using a growing list of verbs may not be a good idea. That said, it is an
```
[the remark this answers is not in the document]
tidoust: Worth exploring, indeed. Ideally, all operations would have some specific meaning defined somewhere (and there would be a restricted number of them). Now, on top of core operations, some of these verbs are the beginning of a named algorithm that the step calls. Such verbs don't need to be defined in Infra. The extraction logic is not smart enough to detect that for now.

src/browserlib/extract-algorithms.mjs
tidoust: Ah yes, I had forgotten that function existed ;)

src/browserlib/extract-algorithms.mjs
```diff
@@ -0,0 +1,649 @@
...
+  const relativeToAbsolute = {};
+  const page = el.closest('[data-reffy-page]')?.getAttribute('data-reffy-page');
+  for (const linkEl of el.querySelectorAll(relativeUrlSelector)) {
+    const attr = linkEl.getAttribute('href') ? 'href' : 'src';
+    const url = new URL(page ?? window.location.href);
+    url.hash = linkEl.getAttribute(attr);
+    relativeToAbsolute[linkEl.getAttribute(attr)] = url.toString();
+  }
+
+  const clone = cloneAndClean(el);
```
dontcallmedom, line 338: to help (somewhat) reduce the weight of the extracts, I wonder if we shouldn't apply the same [cleaning as definition content](https://github.com/w3c/reffy/blob/113e87493ccfe96e39498cb31b686d5fcf5f4a4e/src/browserlib/extract-dfns.mjs#L148-L181) (and thus migrate that code into the cloneAndClean module)

tidoust: That would reduce the weight from ~48MB to ~32MB.

I had considered it but since I did not really know how to scope the analyses we would end up doing against algorithms, I thought I'd preserve the whole thing. With definitions, we want to extract prose that can be inserted somewhere else. We need to be conservative. For analysis, the more the merrier.

There are algorithm steps that contain formulas with syntax highlighting that end up being represented through a `<c->` element, others such as Web IDL that use `<emu-val>` and `<emu-const>`, and some specs seem to integrate recent changes through `<ins>` such as [nav-speculation](https://wicg.github.io/nav-speculation/speculation-rules.html#content-security-policy-patches-match-element-to-source-list). Dropping the elements would cut words in the middle of a sentence. The ["fingerprint" image](https://fetch.spec.whatwg.org/#resolve-an-origin) used sometimes could be worth preserving as well for study.

Now, the vast majority of gains seems to come from cutting attributes. Perhaps we can just do that, but preserve elements for now.

(Also, note there's a bug in the extract-dfns function: `el.closest('[data-reffy-page]')` is unlikely going to return an element, because `el` is a detached clone. I'm surprised that does not seem to have triggered any issue so far. I'll need to fix that, as done in the extract-algorithms module)

dontcallmedom (approved):

dontcallmedom (changes requested):
noted a few bugs while exploring, will suggest changes or discussion in a bit

dontcallmedom (commented):
changes made - the second one in particular deserves review

dontcallmedom:
https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#duration-time-component is extracted as an algorithm, which it isn't really; not sure yet what we can do something about it

src/browserlib/extract-algorithms.mjs
```diff
@@ -613,9 +613,20 @@ function findAlgorithms(section, { includeIgnored } = { includeIgnored: false })
...
+  // Probable one-step algorithms starts with "To "
+  // followed by an exported definition
+  // of dfn-type either "dfn" or "abstract-op"
+  const candidateDfnSelectors = ['dfn[data-export][data-dfn-type="dfn"]',
+				 'dfn[data-export][data-dfn-type="abstract-op"]'
+				];
+  const probableOneLine = [...section.querySelectorAll(candidateDfnSelectors.map(s => `p:has(${s})`).join(','))]
+    .filter(p => p.textContent.startsWith("To " + p.querySelector(candidateDfnSelectors.join(',')).textContent))
+    .map(p => {
+      return { rationale: '"To <dfn>"', root: p };
```
tidoust, line 625: Isn't that going to extract the introductory paragraph of multi-step algorithms too, which will also be extracted with an `<ol>` as root?

```html
<p>To <dfn ...>do plenty of things</dfn>, run the following steps:</p>
<ol><li>...</li></ol>
```

The duplicate filtering some lines below won't catch that because the `<p>` and `<ol>` elements are disjoint.

tidoust:
> https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#duration-time-component is extracted as an algorithm, which it isn't really; not sure yet what we can do something about it

That's precisely the sort of steps that were ignored when the code was only checking the first step in a list. But then extraction missed a number of real algorithms as a result. There's room for improvement, for sure.

tidoust:
I mentioned updating the README, but that's more for Webref in practice. Now, a couple of additional things to do before merge:
1. `src/browserlib/reffy.json` needs to be completed with the new module, otherwise it won't run by default. I'll do it.
2. I noticed that the crawl crashes partially on a few specs (including Web Audio API, WebXR, and Web Authentication), falling back to a regular network request instead of reusing the information cache. Error is "Body is unusable". I need to look into that.

tests/extract-algorithms.js
```diff
@@ -56,6 +56,24 @@ const tests = [
...
+        <p>To <dfn data-export data-dfn-type=dfn id="algo-id">do nothing</dfn>, keep calm and carry on.</p>
+      </section>`,
+    algorithms: [
+      {
+        name: 'do nothing',
+        href: 'about:blank#algo-id',
+        html: 'To <dfn data-export="" data-dfn-type="dfn" id="algo-id">do nothing</dfn>, keep calm and carry on.',
+        rationale: '"To <dfn>"',
+        steps: [ {
+	  html: 'To <dfn data-export="" data-dfn-type="dfn" id="algo-id">do nothing</dfn>, keep calm and carry on.'
```
tidoust, line 72: Conceptually, do we really want to duplicate the text both in the root `html` property, and as a single step? Wouldn't it be clearer to not have any `steps` property in such cases? I.e., to end up with:

```js
    [
      {
        name: 'do nothing',
        href: 'about:blank#algo-id',
        html: 'To <dfn data-export="" data-dfn-type="dfn" id="algo-id">do nothing</dfn>, keep calm and carry on.',
        rationale: '"To <dfn>"'
     }
    ]
```

src/browserlib/extract-algorithms.mjs
dontcallmedom: I had (mistakenly) assumed that the duplication filter would handle it (or that the test would yell at me otherwise); will look into fixing it

tests/extract-algorithms.js
dontcallmedom: I wondered about it, but started with the smallest code change. I think the options are:
* split the actual procedural stuff ("keep calm and carry on" into the step (ideal, but maybe hard to do robustly)
* not have steps (which at least provide a simple signal this is a "short form" algorithm)
* duplicate root markup in a single step

I'm happy to use the 2nd option at least as a starting point, which I agree would be an improvement.

tidoust: +1 for the second option. It  has the advantage of being consistent with the way a step in an algorithm is extracted when it does not define substeps: It gets reported with an `html` property and no `steps`.

dontcallmedom: done in [8487c8e](https://github.com/w3c/reffy/pull/1614/commits/8487c8e5a2902bbdee40dd535daa101381516c43)

src/browserlib/extract-algorithms.mjs
dontcallmedom: done in [ee10949](https://github.com/w3c/reffy/pull/1614/commits/ee109493f3fdb8d1c7b82200f5d31350d7d20027)

dontcallmedom (approved):

dontcallmedom:
feel free to merge if you're satisfied with my changes to address your comments

src/browserlib/extract-algorithms.mjs
```diff
@@ -662,5 +666,9 @@ export default function (spec, idToHeading = {}) {
     return [];
   }
   const algorithms = findAlgorithms(document);
-  return algorithms.map(algo => serializeAlgorithm(algo));
+  return algorithms.map(algo => serializeAlgorithm(algo))
+    // remove duplicate from multi-steps algorithms also detected by the
+    // short form algorithms finder (relying on them being later in the
+    // list)
+    .filter((algo, idx, arr) => arr.findIndex(al => !algo.html || al.html === algo.html) === idx);
```
tidoust, line 673: This assumes that different algorithms will have different `html` properties. That's not the case, that property may contain pretty generic prose. E.g., 
[`createMediaKeys()`](https://w3c.github.io/encrypted-media/#dom-mediakeysystemaccess-createmediakeys) and [`createSession()`](https://w3c.github.io/encrypted-media/#dom-mediakeys-createsession) both end up with "When this method is invoked, the user agent MUST run the following steps". This code would drop the second one as a duplicate of the first.

At a minimum, you might want to only filter when `algo.rationale === 'To <dfn>'`, which should indeed drop the algorithm if it has already been extracted otherwise.

If you do that, I suspect you don't need the differentiated sort, and may be able to move the filtering back into findAlgorithms.

src/browserlib/extract-algorithms.mjs
```diff
@@ -494,24 +494,28 @@ function serializeSteps(root) {
...
+      return [serializeStep(root, level + 1)];
+    }
   }
 }
 
 /**
  * Serialize an algorithm step
  */
-function serializeStep(li) {
+function serializeStep(li, level) {
```
tidoust, line 518: The level parameter is meant for later usage and consistency? Or is there something missing?

src/browserlib/extract-algorithms.mjs
dontcallmedom: I haven't found a way to find equivalence of algorithms at the `findAlgorithms` stage because `.html` isn't populated at that time. Thanks for flagging the fact that `.html` equality isn't sufficient for filtering; rather than `.rationale`, I think I'll make it applicable only if the list of steps is empty (at which points `.html` is guaranteed to be the only meaningful content)

src/browserlib/extract-algorithms.mjs
dontcallmedom: mostly for consistency, with the idea that it would likely be useful later; but I'm happy to remove it if you feel that's premature

src/browserlib/extract-algorithms.mjs
dontcallmedom: done in [11ea7a0](https://github.com/w3c/reffy/pull/1614/commits/11ea7a0a2da7fda11ac159c31546b3d5b86f59bc)

src/browserlib/extract-algorithms.mjs
tidoust: Re-reading this, I realized the switch case serialization was convoluted. The case is now serialized directly at the `case` level.

I dropped the `level` parameter everywhere, because it was no longer needed. We can re-introduce it as needed if the level starts to be relevant.

src/browserlib/extract-algorithms.mjs
dontcallmedom: A clear improvement, thanks!

src/browserlib/extract-algorithms.mjs
```diff
@@ -413,6 +432,10 @@ function getAlgorithmInfo(algo, context) {
     if (container && !context?.nested) {
       if (container.getAttribute('data-algorithm')) {
         info.name = normalize(container.getAttribute('data-algorithm'));
+        if (container.getAttribute('data-algorithm-for')) {
```
dontcallmedom, line 435: Is that convention documented anywhere? I didn't know about it

src/browserlib/extract-algorithms.mjs
```diff
@@ -389,10 +389,29 @@ function getDefinedNameIn(el) {
...
   }
+
+  if (!paragraph) {
+    // Consider that the introductory paragraph is the previous paragraph.
+    // That's not going to be 100% correct. For example, we will incorrectly
+    // capture an intermediary paragraph as in:
+    // https://w3c.github.io/webappsec-csp/#abstract-opdef-parse-a-serialized-csp
+    // TODO: improve!
+    paragraph = algo.root;
+    while (paragraph && paragraph.nodeName !== 'P') {
```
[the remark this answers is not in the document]
tidoust: Many introductory paragraphs don't have `dfn` though, and skipping over paragraphs without definitions could mean extracting a paragraph that does have a definition but is way before the algorithm and has nothing to do with it.

In practice, the CSP spec is the only example that remains (and that I spotted) that creates two algorithms for now. That seems ok enough for now!

src/browserlib/extract-algorithms.mjs
tidoust: Not that I'm aware of. Is `data-algorithm` documented anywhere in practice? Bikeshed and ReSpec talk about the `algorithm` class and attribute, not about the resulting markup.

tidoust:
I made a few adjustments, with related tests. The single-step algorithms seem to get correctly extracted without creating duplicates now (with the CSP exception, which I'm happy to live with for now). Good to merge?

dontcallmedom (approved):
congrats again for an amazing source of yak wool!

PR #1628, opened 2024-07-23
Switch to ECMAScript modules

tidoust:
Shaving yaks... I naively thought a couple of updates would be all that was required to switch to ECMAScript modules. The number of updates grew a bit out of control, but nothing fantastic in there, just turning calls to `require` into calls to `import` and adjusting the logic here and there (e.g., no more `__dirname`). We may still end up with a couple of bad surprises on things that are not fully covered by tests but they should be easy to fix.

I refreshed a bit the post-processor logic to load the modules once at the begining of the crawl so as to avoid making other functions asynchronous.

Scripts retain a `.js` file extension for now, mainly because it was easier to update the code without also changing the extensions. The switch to ECMAScript modules is done in `package.json` instead.

Note the check `process.argv[1] === fileURLToPath(import.meta.url)` is not a perfect replacement for `require.main === module` as it does not account for cases where the script is run without specifying the file extension, but that should still be good enough for the few modules where we still need that.

One **breaking change**: the `getSchemaValidationFunction`, exported by Reffy and used in Webref, becomes asynchronous.

### DIFF
--- a/src/lib/fetch.js
+++ b/src/lib/fetch.js
@@ -6,7 +6,6 @@
  */
 
 import os from 'node:os';
-import { cwd } from 'node:process';
 import path from 'node:path';
 import baseFetch from 'fetch-filecache-for-crawling';
 import { loadJSON } from './util.js';
--- a/src/lib/fetch.js
+++ b/src/lib/fetch.js
@@ -5,16 +5,15 @@
  * @module finder
  */
 
-const os = require('os');
-const path = require('path');
-const baseFetch = require('fetch-filecache-for-crawling');
+import os from 'node:os';
+import { cwd } from 'node:process';
+import path from 'node:path';
+import baseFetch from 'fetch-filecache-for-crawling';
+import { loadJSON } from './util.js';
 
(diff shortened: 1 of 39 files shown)


src/lib/fetch.js
```diff
@@ -5,16 +5,15 @@
  * @module finder
  */
 
-const os = require('os');
-const path = require('path');
-const baseFetch = require('fetch-filecache-for-crawling');
+import os from 'node:os';
+import { cwd } from 'node:process';
```
dontcallmedom, line 9: is that used anywhere?

dontcallmedom (approved):
Thanks a bunch for dealing with that migration! A minor nit, but otherwise looks great to me!

src/lib/fetch.js
tidoust: Nowhere. Good catch! :)

I had initially started loading JSON through dynamic imports and thought I'd need to resolve paths against the current working directory, but then I switched to a common `loadJSON` function (and the current working directory is the base folder for dynamic imports in any case)

PR #1629, opened 2024-07-24
Fix invalid imports and schema validation code

tidoust:
Tests needed for `index.js`, the typos went through the cracks...

Also, the new `loadJSON` function does not throw an error when file cannot be loaded, but rather returns null. The function failed as a result when asked to provide a validation function for a schema that did not exist (Webref tests happilly request a validation function for IDL extracts and only after checks that the extracts are JSON files).


tidoust:
Taking the liberty to merge to release a patch version of Reffy that can actually be imported.

PR #1723, opened 2024-12-09
Extract CDDL definitions

tidoust:
Needed for https://github.com/w3c/webref/issues/1353. Created as draft pull request as I'm not yet fully convinced by the combined extract, and as the `data-cddl-module` convention hasn't been integrated in Bikeshed yet (see https://github.com/speced/bikeshed/pull/2977).

With this update, Reffy looks for and extracts CDDL content defined in `<pre class="cddl">` block. The logic is vastly similar to the logic used for IDL. Shared code was factored out accordingly.

Something specific about CDDL: on top of generating text extracts, the goal is also to create one extract per CDDL module that the spec defines. To associate a `<pre>` block with one or more CDDL module, the code looks for a possible `data-cddl-module` module, or for module names in the `class` attribute (prefixed by `cddl-` or suffixed by `-cddl`). The former isn't used by any spec but is the envisioned mechanism in Bikeshed to define the association, the latter is the convention currently used in the WebDriver BiDi specification.

When a spec defines modules, CDDL defined in a `<pre>` block with no explicit module annotation is considered to be defined for all modules (not doing so would essentially mean the CDDL would not be defined for any module, which seems weird).

When there is CDDL, the extraction produces:
1. an extract that contains all CDDL definitions: `cddl/[shortname].cddl`
2. one extract per CDDL module: `cddl/[shortname]-[modulename].cddl` (I'm going to assume that no one is ever going to define a module name that would make `[shortname]-[modulename]` collide with the shortname of another spec).

Note: some specs that define CDDL do not flag the `<pre>` blocks in any way (Open Screen Protocol, WebAuthn). Extraction won't work for them for now. Also, there are a couple of places in the WebDriver BiDi spec that use a `<pre class="cddl">` block to *reference* a CDDL construct defined elsewhere. Extraction will happily include these references as well, leading to CDDL extracts that contain invalid CDDL. These need fixing in the specs.

### DIFF
--- a/src/browserlib/extract-cddl.mjs
+++ b/src/browserlib/extract-cddl.mjs
@@ -24,18 +24,18 @@ import trimSpaces from './trim-spaces.mjs';
  *
  * @function
  * @public 
- * @return {Promise} The promise to get a dump of the CDDL definitions per
- *   CDDL module, or an empty array if the spec does not contain any CDDL.
+ * @return {Array} A dump of the CDDL definitions per CDDL module, or an empty
+ * array if the spec does not contain any CDDL.
  */
 export default function () {
     // Specs with CDDL are either recent enough that they all use the same
     // `<pre class="cddl">` convention, or they don't flag CDDL blocks in any
     // way, making it impossible to extract them.
-    const cddlSelector = 'pre.cddl:not(.exclude):not(.extract)';
-    const indexSelector = '#cddl-index';
+    const cddlSelectors = ['pre.cddl:not(.exclude):not(.extract)'];
+    const excludeSelectors = ['#cddl-index'];
 
     // Retrieve all elements that contains CDDL content
-    const cddlEls = getCodeElements([cddlSelector], [indexSelector]);
+    const cddlEls = getCodeElements(cddlSelectors, { excludeSelectors });
 
     // Start by assembling the list of modules
     const modules = {};
--- a/src/browserlib/extract-cddl.mjs
+++ b/src/browserlib/extract-cddl.mjs
@@ -10,8 +10,9 @@ import trimSpaces from './trim-spaces.mjs';
  * 
  * Each CDDL module is represented as an object with the following keys whose
  * values are strings:
- * - shortname: the CDDL module shortname. Shortname is "" if there are no
- * - label: A full name for the CDDL module.
+ * - shortname: the CDDL module shortname. Shortname is "" if the spec does not
+ * define any module, and "all" for the dump of all CDDL definitions.
+ * - label: A full name for the CDDL module, when defined.
  * - cddl: A dump of the CDDL definitions.
  *
  * If the spec defines more than one module, the first item in the array is the
@@ -36,15 +37,16 @@ export default function () {
     // Retrieve all elements that contains CDDL content
     const cddlEls = getCodeElements([cddlSelector], [indexSelector]);
 
-    // By convention, CDDL defined without specifying a module is defined
-    // for all modules (that CDDL would essentially be lost otherwise, there's
-    // no reason for a spec to define CDDL for no module if it uses modules).
-    // Start by assembled the list of modules
+    // Start by assembling the list of modules
     const modules = {};
     for (const el of cddlEls) {
         const elModules = getModules(el);
         for (const name of elModules) {
-            modules[name] = [];
+            // "all" does not create a module on its own, that's the name of
+            // the CDDL module that contains all CDDL definitions.
+            if (name !== 'all') {
+                modules[name] = [];
+            }
         }
     }
 
@@ -55,6 +57,7 @@ export default function () {
         if (!cddl) {
             continue;
         }
+        // All CDDL appears in the "all" module.
         mergedCddl.push(cddl);
         let elModules = getModules(el);
         if (elModules.length === 0) {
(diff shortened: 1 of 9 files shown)


tidoust:
I made some updates on the way modules are processed following discussions in https://github.com/w3c/webref/issues/1353:
- When modules are used, the extract that contains all CDDL definitions is now named `[shortname]-all.cddl`. This is meant to avoid ending up with `[shortname].cddl` that looks like the *right* entry point. That collection of CDDL definitions may well not be directly usable for validation of arbitrary CBOR data items (first rule will likely be for one of the modules).
- Specs can now also define CDDL definitions specifically for the "all" module through a `data-cddl-module="all"` attribute. This makes it possible to define a root rule in the spec. Side note: such CDDL definitions would not appear in the generated CDDL index that Bikeshed would generate if support for CDDL gets added. I think that's fine for now, I don't see any reason to define additional CDDL definitions for the "all" module on top of creating a root rule.

Running a local crawl, I end up with the following list of CDDL extracts:
- `at-driver-all.cddl`
- `at-driver-local.cddl`
- `at-driver-remote.cddl`
- `permissions-all.cddl`
- `permissions-local.cddl`
- `permissions-remote.cddl`
- `web-bluetooth-all.cddl`
- `web-bluetooth-local.cddl`
- `web-bluetooth-remote.cddl`
- `webdriver-bidi-all.cddl`
- `webdriver-bidi-local.cddl`
- `webdriver-bidi-remote.cddl`

CDDL defined in other specs (Open Screen Application Protocol, Open Screen Network Protocol, Web Authentication) does not appear because these specs do not yet flag the `<pre>` tags that contain the CDDL in any way.

The extracted CDDL currently contains occurrences of invalid `EmptyResult` constructs. That's because the specs currently also use a `<pre class="cddl">` tag to *reference* the rule, (e.g., [`session.end`](https://w3c.github.io/webdriver-bidi/#command-session-end)). That's not an extraction bug, specs need to be fixed.

I don't have an immediate solution for validating the CDDL in Webref for now. That would be next step :)

src/browserlib/get-code-elements.mjs
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Helper function that returns a set of code elements in document order based
+ * on a given set of selectors, excluding elements that are within an index.
+ *
+ * The function excludes elements defined in informative sections.
+ *
+ * The code elements are cloned and cleaned before they are returned to strip
+ * annotations and other asides.
+ */
+export default function getCodeElements(codeSelectors, excludeSelectors) {
```
dontcallmedom, line 13: I think the code might be a bit easier to parse if excludeSelectors was part of an option object (à la `getCodeElements(selectors, {excludeSelectors: []})`)

src/browserlib/get-code-elements.mjs
```diff
@@ -0,0 +1,22 @@
...
+    return [...document.querySelectorAll(codeSelectors.join(', '))]
+        // Only keep the elements that are not within the index at the end of
+        // the specification and that are defined in a normative section.
```
dontcallmedom, line 16:
```suggestion
        // Skip excluded and elements and those in informative content
```

src/browserlib/trim-spaces.mjs
```diff
@@ -0,0 +1,36 @@
...
+    const lines = code.trim().split('\n');
+    const toRemove = lines
+        .slice(1)
+        .filter(line => line.search(/\S/) > -1)
+        .reduce(
+            (min, line) => Math.min(min, line.search(/\S/)),
+            Number.MAX_VALUE);
+    return lines
+        .map(line => {
+            let firstRealChat = line.search(/\S/);
```
dontcallmedom, line 24: I know this is a copy from existing code, but was this meant as `char` rather than `chat`?

src/browserlib/extract-cddl.mjs
```diff
@@ -0,0 +1,125 @@
...
+ * If the spec defines more than one module, the first item in the array is the
+ * "all" module that contains a dump of all CDDL definitions, regardless of the
+ * module they are actually defined for (the assumption is that looking at the
+ * union of all CDDL modules defined in a spec will always make sense, and that
+ * a spec will never reuse the same rule name with a different definition for
+ * different CDDL modules).
+ *
+ * @function
+ * @public 
+ * @return {Promise} The promise to get a dump of the CDDL definitions per
```
dontcallmedom, line 27: doesn't seem to be returning a Promise?

dontcallmedom (approved):
LGTM with a few nits

src/browserlib/trim-spaces.mjs
tidoust: Ha ha, yes! I had bumped into it and wondered what `chat` was supposed to mean, and surrendered because the code just worked. It hadn't occurred to me it was just a typo :)

src/browserlib/extract-cddl.mjs
tidoust: Indeed. And that's also a copy-and-paste from IDL extraction (which also incorrectly says it returns a Promise). I guess that shows these comments are not particularly useful since we don't use them. Actual typing would be more useful. But that's a topic for another day. I'll fix the comments in the meantime.

PR #1753, opened 2025-01-20
Extract CSS legacy name aliases defined in specs

### LINKED ISSUE #538: Improve extracts for legacy and aliases CSS properties
See discussion in https://github.com/w3c/webref/pull/115#discussion_r591321416.

Goal is to make sure that all CSS properties that appear in extracts have some value definition, somehow, be it through a `prose` key.

Currently, Reffy just manages to extract the property name for these legacy properties. It would be good if it could be more specific. For instance, for `-webkit-` properties that are aliases, link to the alias, and/or add a `prose` key that gives some info about the property.

tidoust:
CSS specs sometimes define property names that are "legacy name aliases" of another property name. Reffy extracted the legacy property names (during post-processing) as they are defined in a `<dfn>` element but did not flag them in any way.

The crawler now adds a `legacyAliasOf` key to the property in the CSS extract whose value is the name of the aliased property.

Extraction is based on looking for links to the term "legacy name alias" (defined in CSS Cascade 4 and 5) and simple patterns around it. Either:
1. a dfn for a property followed by a reference to the aliased property; or
2. a table with two columns: dfns in the first column, references to the aliased properties in the second column.

First case handles legacy definitions in `css-align-3`, `css-fonts-4`, and `css-ui-4`. Second case handles legacy definitions in `compat`.

Legacy definitions in `css-text-3` and `css-text-4` are not extracted... and that is a good thing because they define `word-wrap` both as an actual property and as a legacy name alias of `overflow-wrap`, which seems wrong.

Legacy definitions in `css-flexbox-1` are not extracted either... which also is a good thing given that `compat` handles them already!

The `css-ui-4` spec also defines `-webkit-user-select` as an alias of `user-select`, but the spec [does not use the "legacy name alias" mechanism](https://drafts.csswg.org/css-ui-4/#propdef--webkit-user-select), possibly on purpose. That relationship will have to be added with a patch if we really want it.

Fixes #538 (although note there will remain a few CSS properties in extracts that do not have a real "value") and https://github.com/w3c/webref/issues/1427

### DIFF
--- a/src/browserlib/extract-cssdfn.mjs
+++ b/src/browserlib/extract-cssdfn.mjs
@@ -18,13 +18,17 @@ export default function () {
   const warnings = [];
 
   const res = {
-    // Properties are always defined in dedicated tables in modern CSS specs
-    properties: extractDfns({
-      selector: 'table.propdef:not(.attrdef)',
-      extractor: extractTableDfns,
-      duplicates: 'merge',
-      warnings
-    }),
+    // Properties are always defined in dedicated tables in modern CSS specs,
+    // Legacy properties are always defined in prose in a dfn with a nearby
+    // reference to "legacy name alias"
+    properties: []
+      .concat(extractDfns({
+        selector: 'table.propdef:not(.attrdef)',
+        extractor: extractTableDfns,
+        duplicates: 'merge',
+        warnings
+      }))
+      .concat(extractLegacyProperties(document)),
 
     // At-rules, selectors, functions and types are defined through dfns with
     // the right "data-dfn-type" attribute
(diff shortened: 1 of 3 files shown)


src/browserlib/extract-cssdfn.mjs
```diff
@@ -826,3 +830,56 @@ const extractProductionRules = root => {
...
+ * Extract legacy alias relationships, looking for occurrences of the term
+ * "legacy name alias".
+ *
+ * Next to it, there should be:
+ * 1. a dfn for a property followed by a reference to the aliased property; or
+ * 2. a table with two columns: dfns in the first column, references to the
+ * aliased properties in the second column.
+ */
+const extractLegacyProperties = doc =>
+  [...doc.querySelectorAll('a[href*="#legacy-name-alias"]')]
```
dontcallmedom, line 845:
```suggestion
  [...doc.querySelectorAll('a[href$="#legacy-name-alias"]')]
```

dontcallmedom (approved):
Nice!

A couple of suggested minor improvements for your consideration

tidoust:
> Legacy definitions in `css-text-3` and `css-text-4` are not extracted... and that is a good thing because they define `word-wrap` both as an actual property and as a legacy name alias of `overflow-wrap`, which seems wrong.

Raised in https://github.com/w3c/csswg-drafts/issues/11544

> Legacy definitions in `css-flexbox-1` are not extracted either... which also is a good thing given that `compat` handles them already!

Raised in https://github.com/w3c/csswg-drafts/issues/11545

PR #1777, opened 2025-02-17
Drop cache info when a redirection took place

### LINKED ISSUE #1774: Crawl results stall when there's a redirect
Via https://github.com/w3c/webref/issues/1468. When a spec gets moved to another group and URL, Reffy receives a redirect response and crawls the resulting spec. However, it uses the cache headers of the redirect response to determine whether the spec further changes, and thus misses actual updates made to the spec.

tidoust:
When a redirection takes place, Reffy follows the redirection logic (typically done through scripting) but the cache info it gets from Puppeteer remains for the initial URL. Reffy incorrectly assumed that info also applied to the final page.

There's no easy way to retrieve the cache info of the final URL. Since that should only affect a spec that moves, and only until we detect and update the URL of the spec in browser-specs, this update simply drops the cache info to force Reffy to crawl the spec.

Fix #1774.

### DIFF
--- a/src/lib/mock-server.js
+++ b/src/lib/mock-server.js
@@ -170,12 +170,39 @@ mockAgent
   });
 
 mockAgent
+  .get("https://www.w3.org")
+  .intercept({ method: "GET", path: "/TR/iredirect/" })
+  .reply(200,
+    `<!DOCTYPE html><script>window.location = '/TR/recentlyupdated/';</script>`,
+    {
+      headers: {
+        "Content-Type": "text/html",
+        "Last-Modified": "Fri, 11 Feb 2022 00:00:42 GMT"
+      }
+    }
+  );
+
+mockAgent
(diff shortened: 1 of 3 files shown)


src/lib/mock-server.js
```diff
@@ -169,13 +169,40 @@ mockAgent
...
+  );
+
 mockAgent
   .get("https://drafts.csswg.org")
   .intercept({ method: "GET", path: "/server-hiccup/" })
   .reply(200,
     `<html><title>Server hiccup</title>
     <h1> Index of Server Hiccup Module Level 42 </h1>`,
-    { header: { "Content-Type": "text/html" } })
+    { headers: { "Content-Type": "text/html" } })
```
tidoust, line 205: Unrelated fix. I don't think this had any practical implication in practice, but that was still wrong...

dontcallmedom (approved):

PR #1789, opened 2025-03-06
Dfns extraction: expand logic that skips imported definitions

### LINKED ISSUE #1788: Dfns extraction: expand logic that skips imported definitions
Via https://github.com/tc39/ecma426/issues/174

The crawler skips "imported definitions", meaning those that some specs give a `<dfn>` to, but that are just a proxy to a definition in some other spec. That logic expects the markup of these imported definitions to be `<dfn><a>foo</a></dfn>`. The Ecmarkup tool seems to generate the opposite pattern: `<a><dfn>foo</dfn></a>`, which the crawler does not yet handle.

tidoust:
The crawler expected "imported definitions" to use `<dfn>` as the wrapping element with an inner `<a>` element. The Source map format spec now uses the opposite pattern. This update expands the logic accordingly.

Close #1788.

### DIFF
--- a/src/browserlib/extract-dfns.mjs
+++ b/src/browserlib/extract-dfns.mjs
@@ -341,7 +341,9 @@ export default function (spec, idToHeading = {}) {
     // (pending a proper dfns curation process, see:
     // https://github.com/w3c/webref/issues/789)
     .filter(node => {
-      const link = node.querySelector('a[href^="http"]');
+      const link =
+        node.querySelector('a[href^="http"]') ??
+        node.closest('a[href^="http"]');
       return !link ||
         (node.textContent.trim() !== link.textContent.trim()) ||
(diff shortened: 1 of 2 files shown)


dontcallmedom:
since the upstream spec has been changed, I wonder if it's still worth merging this? did you check if it would generate true and/or false positives?

tidoust:
No false positives generated! Don't you trust our test suite? 😲

And it's actually useful as it captures two cases of that same pattern used in other specs:
- [default reading order](https://w3c.github.io/audiobooks/#dfn-default-reading-order) in Audiobooks
- [lexical form](https://w3c.github.io/json-ld-api/#dfn-lexical-form) in the JSON-LD API

dontcallmedom (approved):

PR #1841, opened 2025-05-21
Extract alternate ids for headings

### LINKED ISSUE #1824: Extract heading ids from <hN> tags in addition to their <section>s
Respec is generating `<section>` elements with one `id` that wrap `<hN>` tags with a different `id`. You can see this at https://w3ctag.github.io/privacy-principles/#hl-sensitive-information, where that anchor points to an `<h3 id="hl-sensitive-information">`, which is surrounded by a `<section id="sensitive-information">`. The generated self-link on the heading goes to `#hl-sensitive-information`, but https://w3c.github.io/webref/ed/headings/privacy-principles.json only includes `"id": "sensitive-information"`. I think Respec is doing the wrong thing here, and I'll file a bug with them too, but Reffy should probably handle this kind of weird markup and include both `id`s.

dontcallmedom:
fixes #1824

### DIFF
--- a/src/browserlib/map-ids-to-headings.mjs
+++ b/src/browserlib/map-ids-to-headings.mjs
@@ -86,21 +86,21 @@ export default function () {
     let href = nodeid;
 
     if (parentSection) {
-      let alternateIds = [];
+      const ids = [];
       let id;
 
       const heading = parentSection.heading;
       if (heading.id) {
         id = heading.id;
         href = getAbsoluteUrl(heading, { singlePage });
-	alternateIds.push(id);
+	ids.push(id);
       }
       else {
         const anchor = heading.querySelector('a[name]');
         if (anchor) {
           id = anchor.getAttribute('name');
           href = getAbsoluteUrl(anchor, { singlePage, attribute: 'name' });
-	  alternateIds.push(id);
+	  ids.push(id);
         }
       }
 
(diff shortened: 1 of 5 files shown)


src/browserlib/map-ids-to-headings.mjs
```diff
@@ -114,8 +117,12 @@ export default function () {
       if (id) {
         mapping.id = id;
       }
+      alternateIds = alternateIds.filter(id => id !== mapping.id);
```
tidoust, line 120: Instead of filtering out, I would find it more natural to have `alternateIds` really be the list of alternate IDs from the start. Unless I missed something, that seems doable if you:
- drop the `push` lines 96 and 103
- add something like `if (id) { alternateIds.push(id); }` line 108 before overriding `id`
- drop this line 120

Alternatively, the variable could be called `ids`. You'd push all ids, you'd pop the last one for use as `mapping.id`, and set the other one as `mapping.alternateIds` if there's a rest. Up to you :)

dontcallmedom: I'm not sure I understood if you meant the final data should always have all ids, or if this was only about code structure; I've updated the code on the basis of the latter for now

tidoust: I confirm I meant the latter.

PR #1849, opened 2025-05-30
Add cssmerge post-processor to consolidate CSS extracts

tidoust:
This adds a `cssmerge` post-processor at the crawl level that consolidates definitions from CSS extracts into a single file. The CSS entries in the resulting file are de-duplicated. The module starts with a number of comments that detail the approach.

The intent is to to address https://github.com/w3c/webref/issues/1519 by creating a view similar to that offered in MDN data.

Comments at the top of the module also detail the main differences between the two views. Main one is structural: we tend to use arrays in Webref so this code also produces arrays, while MDN data uses indexed objects. It's easy to build an indexed view from an array though.

Consolidation focuses on syntaxes and the code does not attempt to compute the `units` list that exists in MDN data. It is easy to build that list by looking at values of a handful of CSS properties. We could add it later on to the consolidated file if that proves needed.

To de-duplicate, the most recent definition is used. That's the only approach I can think of that can be automated. Consumers of MDN data may need a more nuanced approach where the syntax better matches what core browsers currently support. We can perhaps refine the process later on if needed with some sort of patching mechanism (but my hunch is that this will remain a manual process).

The resulting file can be seen in [tidoust/mdn-webref](https://github.com/tidoust/mdn-webref/?tab=readme-ov-file#comparing-data-in-mdn-data-and-webref). That repository also lists differences that exist between Webref, MDN data, and syntax patches defined in CSSTree. Some of these differences require updating the specs. Other differences are the result of adopting a live-on-the-edge approach as mentioned above while MDN data is more conservative.

I'm creating the pull request as a draft. On top of possibly discussing "liberties" that the code takes with the definitions, I think this needs:

- [ ] A few **tests**. I'll take care of that.
- [ ] A **publication strategy**. That is, I do not know how we're going to publish the resulting file in an npm package. Should we push the file in `@webref/css`? If so, how do we make it clear that the package contains both CSS extracts and the consolidated version of these extracts? (Among other things, the package already contains a root `CSS.json` file). Or should we publish another package? We may need to change the approach (or the name of the file that gets generated) depending on that.
- [ ] Review of the resulting structure and data by interested parties. Does it provide a basis for syntaxes that is sound enough for MDN data and others?

### DIFF
--- a/src/lib/post-processor.js
+++ b/src/lib/post-processor.js
@@ -52,6 +52,7 @@ import path from 'node:path';
 import { pathToFileURL } from 'node:url';
 import { createFolderIfNeeded, shouldSaveToFile } from './util.js';
 import csscomplete from '../postprocessing/csscomplete.js';
+import cssmerge from '../postprocessing/cssmerge.js';
 import events from '../postprocessing/events.js';
 import idlnames from '../postprocessing/idlnames.js';
 import idlparsed from '../postprocessing/idlparsed.js';
@@ -64,6 +65,7 @@ import patchdfns from '../postprocessing/patch-dfns.js';
  */
 const modules = {
   csscomplete,
+  cssmerge,
   events,
   idlnames,
   idlparsed,
--- a/src/postprocessing/cssmerge.js
+++ b/src/postprocessing/cssmerge.js
@@ -1,0 +1,321 @@
+/**
+ * Post-processing module that consolidates CSS extracts into a single
+ * structure. That structure is an object whose keys are `atrules`,
+ * `functions`, `properties`, `selectors`, and `types`. Values are lists of CSS
+ * constructs whose type matches the key.
+ *
+ * CSS constructs follow the same structure as that in individual CSS extracts
+ * except that values that are listed under `values` in the CSS extracts are
+ * not reported in the resulting structure because these values are a mix bag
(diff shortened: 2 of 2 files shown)


tidoust:
Summarizing next steps following discussion:
1. Merge this pull request and release a version of Reffy that can consolidate CSS extracts
2. Start generating the consolidated `css.json` file in Webref
3. Point a few people at the consolidated file (and gap analysis) to validate the approach and identify potential blockers.
4. In parallel with 3, prepare code and start releasing two major versions of `@webref/css` in parallel: v6 with all extracts, v7 only with the consolidated file. v6 will continue to be maintained for a few months at least (that should not add much work to our review process). We may need to introduce breaking changes in v7 based on feedback, it seems a good idea to flag the package as being a pre-release version initially. That could be done by appending a `-alpha` suffix. I believe tools such as dependabot understand that pattern. I don't know yet whether they ignore the pre-release version by default or still suggest updating the dependency.

One open question is whether reviewing the consolidated file diff in Webref is going to be the right approach to check updates before we release new versions of the v7 package. That's to be determined. From a semantic perspective, it should make sense. Getting back to the spec that triggered a specific update may be slightly more involved for features defined and extended in multiple specs, but there aren't so many of them in practice.

PR #1875, opened 2025-07-07
Add anchors for web developers to dfns extracts

tidoust:
For context, see discussion starting at:
https://github.com/mdn/browser-compat-data/pull/23958#issuecomment-3035921128

Some specs such as DOM, Encoding, HTML contain sections targeted at web developers. These sections re-define terms normatively defined elsewhere in a more developer-friendly way. Terms re-defined in these sections are good targets for documentation but did not appear in definitions extracts.

This update makes Reffy parse "for web developers" sections and extract the definitions they contain. This is a prerequisite to publishing a package with definitions that could be used to validate URLs in BCD and web-features, as envisioned in:
https://github.com/w3c/webref/issues/1198#issuecomment-3032925937

Worth noting:
- Ideally, spec authoring tools would provide better support for this pattern, giving definitions more stable IDs than `ref-for-[foo][number]` and creating proper dfns themselves. If they do that, the custom processing introduced here would become moot. Going through tools and specs will take time though. The custom processing done here allows to add the definitions right away. It does not solve the "unstable" IDs issue, but at least provides a theoretical way to identify situations where the ID of a dev dfn changes.
- To keep the cross-references database useful, newly extracted definitions need to be in a separate dfn namespace, i.e., have their own dfn type. Problem is that they also have a "natural" dfn type such as `interface`, `method` or `attribute`. The solution implemented here is to prefix their type with `dev-`. That duplicates dfn types. A cleaner solution would record the "dev" namespace in another property. But that would surprise spec authoring tools. An alternative approach would be to give all of these dfns a `dev` dfn type, but we'd then lose information that could turn out to be useful.
- The key marker for sections targeted at web developers is the use of a `domintro` class. Now, a few specs do use `domintro` in normative definition lists (shape-detection-api, image-capture, mediastream-recording). That's probably unintentional. I'll look into fixing the specs. The code skips `domintro` sections that look suspicious.
- This would add **2815 definitions** to the dfns extracts (which currently contain ~50000 definitions)

### DIFF
--- a/schemas/browserlib/extract-dfns.json
+++ b/schemas/browserlib/extract-dfns.json
@@ -24,20 +24,27 @@
         "enum": [
           "property", "descriptor", "value", "type",
           "at-rule", "function", "selector",
-
           "namespace", "interface", "constructor", "method", "argument",
           "attribute", "callback", "dictionary", "dict-member", "enum",
           "enum-value", "exception", "const", "typedef", "stringifier",
           "serializer", "iterator", "maplike", "setlike", "extended-attribute",
           "event", "permission",
-
           "element", "element-state", "element-attr", "attr-value",
-
           "cddl-module", "cddl-type", "cddl-parameter", "cddl-key", "cddl-value",
-
           "scheme", "http-header",
+          "grammar", "abstract-op", "dfn",
 
-          "grammar", "abstract-op", "dfn"
+          "dev-property", "dev-descriptor", "dev-value", "dev-type",
+          "dev-at-rule", "dev-function", "dev-selector",
+          "dev-namespace", "dev-interface", "dev-constructor", "dev-method", "dev-argument",
+          "dev-attribute", "dev-callback", "dev-dictionary", "dev-dict-member", "dev-enum",
+          "dev-enum-value", "dev-exception", "dev-const", "dev-typedef", "dev-stringifier",
+          "dev-serializer", "dev-iterator", "dev-maplike", "dev-setlike", "dev-extended-attribute",
+          "dev-event", "dev-permission",
+          "dev-element", "dev-element-state", "dev-element-attr", "dev-attr-value",
+          "dev-cddl-module", "dev-cddl-type", "dev-cddl-parameter", "dev-cddl-key", "dev-cddl-value",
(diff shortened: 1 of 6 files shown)


tidoust:
> I'm not a big fan of duplicating the whole definition namespace, and that feels more likely to create challenges over time (given they'll necessarily be duplicate names when considered without namespace) for editing tools.

The `type` is the only property that creates "hard" namespaces in editing tools (no way to target an `interface` definition with `[=Foo=]` for example, you need to use `{{Foo}}`).


> I'm not sure I understood what you meant with
> 
> > A cleaner solution would record the "dev" namespace in another property. But that would surprise spec authoring tools.
> 
> My personal intuition would be to use e.g. a new "devhref" property (or maybe more extensibly start a "links" array of objects with an `href` and a `type`, which would initially only have "dev" as a potential type)

By "dev" namespace, I was thinking of adding a property such as `"audience": "dev"` (or whatever name we may want to use) to the entry.

Now, my working assumption was that we would use separate entries for the dev definitions. They mirror other definitions, but they also seem to be definitions on their own, and could perhaps end up being defined with proper `<dfn>` tags. Until that day comes, it indeed seems a better idea not to create entries and to rather record the URLs within a `links` property in the base definition. I'll look into it.

tidoust:
I updated the approach to create a `links` property instead. On top of the URL itself, I'm gathering contextual info so that these anchors can be used for more than validation of references. That includes the text content of the enclosing `<dt>` to offer a way to distinguish between links when there's more than one (for example [document.fullscreenElement](https://fullscreen.spec.whatwg.org/#ref-for-dom-document-fullscreenelement%E2%91%A0) and [shadowRoot.fullscreenElement](https://fullscreen.spec.whatwg.org/#ref-for-dom-document-fullscreenelement%E2%91%A1) in Fullscreen).

One potential caveat with this approach is noted in the commit message: "Some anchors for developers target definitions in external specs. Previous code ignored them but could have added them as separate entry if need had arisen. There is no easy way to add these anchors with the new approach if they turned out to be useful".

dontcallmedom (approved):
thanks!

PR #1886, opened 2025-07-15
Drop "Appendix" prefix from heading number

tidoust:
We expect heading "numbers" to be actual numbers or letters but appendix sections in RFCs typically start with "Appendix", e.g.: https://www.rfc-editor.org/rfc/rfc8610#appendix-A

Headings extraction produced "Appendix A" as heading number, which fails schema validation. This update drops the "Appendix" prefix to get back to "A".

### DIFF
--- a/src/browserlib/extract-headings.mjs
+++ b/src/browserlib/extract-headings.mjs
@@ -33,7 +33,7 @@ export default function (spec, idToHeading) {
         .replace(/^\s*\./, '')
         .trim(),
       level: headingLevel,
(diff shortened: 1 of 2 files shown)


dontcallmedom (approved):
LGTM in terms of alignment.

I guess this made me realize we lost information in our heading extractions, i.e. in my virtual use case of "regenerate ToC from `headings.json`", we wouldn't be able to prefix them as appendices, but that can wait until that use case becomes less virtual :)

Issue #1890, opened 2025-07-18
Extract CSS definitions in heading elements


cdoublev:
Follow-up on https://github.com/w3c/csswg-drafts/issues/12458.

Reffy is missing `<whole-value>` definition because it uses a heading element, which is not extracted from the crawled content, if I am not mistaken.

Would it be possible to add support for that? :smile_cat:

tidoust:
I adjusted the extraction logic for CSS definitions to better align with the definitions logic.

The `<whole-value>` type is now correctly reported, see:
https://github.com/w3c/webref/blob/main/ed/css/css-values-5.json#L551

Issue #1878, opened 2025-07-07
Extraction of the CSS `<boolean-expr[]>` type


cdoublev:
In the CSS grammar, the [boolean expression multiplier](https://drafts.csswg.org/css-values-5/#boolean), `<boolean-expr[]>`, is like a "template production": other syntaxes can reference it by passing a syntax inside `[]`, referenced as `<test>` in the right hand side of the template production.

Currently, the TR version of CSS Values 5 defines it as `<boolean-expr[]>`, but the ED version also defines it as `<boolean-expr>`, which seems intentional based on [this commit](https://github.com/w3c/csswg-drafts/commit/e7f9e68e1429bae56955570be76bcafb9169b504), but is now inconsistent... 

Neither links to the corresponding production rule, which `reffy` fails to extract because the [regular expression](https://github.com/w3c/reffy/blob/0036e846424543c45ffd65b9abf4c164dd9ee3a7/src/browserlib/extract-cssdfn.mjs#L614) to split productions in the same `<pre>` is inappropriate. (I used `/\s+(?=(<[^>]+>)|([az-]+\(\)))\s*=/i` with success, fwiw.)

From the perspective of implementing a CSS parser, it would be simpler to get `<boolean-expr>` or `<boolean-expr[]>` as the left hand side (the type name) rather than `<boolean-expr[<test>]>`.

In the meantime, I suspect the CSSWG will not define another expression mutliplier anytime soon. So I am not sure it is worth updating `reffy` and possibly the spec.

---

Nit: while debugging in `src/browserlib/extract-cssdfn.mjs`, I realized that this change would avoid multiple passes on the same production rules:

```diff
  const rules = [];
  [...root.querySelectorAll('pre.prod:not(:has(del)):not(:has(ins))')]
-   .concat([...root.querySelectorAll('pre:not(.idl):not(:has(.idl)):not(:has(del)):not(:has(ins))')]
+   .concat([...root.querySelectorAll('pre:not(.prod):not(.idl):not(:has(.idl)):not(:has(del)):not(:has(ins))')]
```

PR #1891, opened 2025-07-21
[CSS extracts] Handle headings, multi dfns, and `<boolean-expr[]>`

tidoust:
The `css-values-5` spec uses new patterns that were not supported yet:

- Definitions in headings, see #1890. Such definitions were correctly captured in the dfns extracts but CSS extraction missed them.
- Definitions that define more than one type at once through `data-lt` attributes, as discussed in https://github.com/w3c/csswg-drafts/pull/12349. More generally speaking, CSS extraction did not always take the `data-lt` attribute into account.
- Definition of the `<boolean-expr[]>` notation, see #1878. The regular expression used to split production rules wasn't expecting spaces in the left-hand side. Note extraction drops `[ <test> ]` and ends up with `<boolean-expr>` as type name. That's not necessary ideal, but there's no good way to represent that type in the extracts. Consumers need dedicated logic to handle the notation.

The update also simplifies the selectors used to find production rules in a document to make them more readable, and to avoid processing the production rules more than once, as discussed in #1878.

Note: for `<boolean-expr[]>`, the extraction will miss the [`<boolean-expr-group>` type](https://drafts.csswg.org/css-values-5/#boolean) that the spec mentions but does not define as a type. That's normal: a `<dfn>` is needed for type extraction to work (for once, we guarantee the presence of an `href` key for all types, and there wouldn't be any). The type will have to be handled in Webref through patching.

### DIFF
--- a/src/browserlib/extract-cssdfn.mjs
+++ b/src/browserlib/extract-cssdfn.mjs
@@ -17,6 +17,8 @@ export default function () {
   // sense of the CSS definitions and production rules it contains
   const warnings = [];
 
+  const dfnElSelector = ':is(dfn,h2,h3,h4,h5,h6)';
+
   const res = {
     // Properties are always defined in dedicated tables in modern CSS specs,
     // Legacy properties are always defined in prose in a dfn with a nearby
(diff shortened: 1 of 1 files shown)


src/browserlib/extract-cssdfn.mjs
```diff
@@ -35,26 +35,26 @@ export default function () {
     // Note some selectors are re-defined locally in HTML and Fullscreen. We
     // won't import them.
     atrules: extractDfns({
-      selector: 'dfn[data-dfn-type=at-rule]:not([data-dfn-for])',
-      extractor: extractTypedDfn,
+      selector: ':is(dfn,h2,h3,h4,h5,h6)[data-dfn-type=at-rule]:not([data-dfn-for])',
```
dontcallmedom, line 38: maybe DRY `:is(dfn,h2,h3,h4,h5,h6)` in a constant?

dontcallmedom (approved):
LGTM, although I won't pretend I parsed all the regexps :)

a minor suggestion within

PR #1901, opened 2025-07-24
Fix css schema

tidoust:


tidoust:
Rha, didn't start from main, not the right PR :(

PR #1926, opened 2025-08-28
Expand JSON schema of idlparsed extracts

tidoust:
Expansion is the result of a mix of:
- copying descriptions from the parsing code;
- completing missing object members leveraging the schema built with AI tools as part of Geek week exploration;
- re-reading the schema.

I validated the new schema against actual data in Webref.

### DIFF
--- a/schemas/common.json
+++ b/schemas/common.json
@@ -800,9 +800,9 @@
               "const": "number"
             },
             "value": {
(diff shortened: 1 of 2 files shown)


dontcallmedom (commented):
some suggested refinements (not tested against data though)

PR #1984, opened 2025-11-26
[Events] Simplify "not bubbling" logic and fix missing targets

### LINKED ISSUE #1983: [Events] Consolidation drops non tree root interfaces when bubbles is not set
Some events that are not explicitly defined as bubbling or non bubbling target multiple interfaces in a bubbling tree, e.g., `Window` and `HTMLElement`. A quick analysis shows that this is the case for: [`blur`](https://html.spec.whatwg.org/multipage/indices.html#event-blur), [`clipboardchange`](https://w3c.github.io/clipboard-apis/#eventdef-globaleventhandlers-clipboardchange), [`error`](https://html.spec.whatwg.org/multipage/indices.html#event-error) (indirectly, because the event targets the `GlobalEventHandlers` mixin, and [`focus`](https://html.spec.whatwg.org/multipage/indices.html#event-focus).

When the spec is not explicit about bubbling, the [consolidation logic](https://github.com/w3c/reffy/blob/main/src/postprocessing/events.js#L134-L149) only sets bubbling to `false` when the root interface of the (potentially) bubbling tree is not in the list of targets. I suspect we did that to make sure we wouldn't claim that an event does not bubble when we weren't sure that was the intent.

Now, when there's no indication of whether the event bubbles or not, the consolidation logic basically ignores non root interfaces, because there's no `else` clause to `if (event.hasOwnProperty('bubbles'))`. That seems like a bug. Testing locally, adding the `else` clause would fix the definitions for the events mentioned above (and not change anything else).

Note: for `clipboardchange`, that would create additional targets on top of `Window`, while the spec only fires the event at the global... but that's a direct consequence of the event being defined as targeting `GlobalEventHandlers`, and the only way to fix that given our extraction logic would be through an event patch.

Instead of adding an `else` clause, I'm wondering whether the consolidation logic could actually set a `bubbles: false` property for events that target interfaces in a bubbling tree. On top of the above events, this would also add `bubbles: false` properties to [`gamepadconnected`](https://w3c.github.io/gamepad/#dfn-gamepadconnected) and [`gamepaddisconnected`](https://w3c.github.io/gamepad/#dfn-gamepaddisconnected)... And now that I look into these gamepad events, I think they also only fire at the global interface in practice, which means they should probably be patched as well to drop other interfaces that include the `WindowEventHandlers` mixin. In any case, this does not seem to create incorrect data.

Am I missing something that would make that a bad idea?

tidoust:
This fixes #1983. We had been erring on the safe side when consolidating events by only flagging events that fire at (non root) interfaces in a tree as non bubbling when the event did not fire at the root interface. That approach works but there was a bug that made consolidation forget non root interfaces when an event did not have a `bubbles` flag.

In practice, in the absence of explicit information, it seems fine to flag all events that fire at non root interfaces in a tree as non bubbling. This update simplifies the consolidation logic to always do that. That simplification fixes the underlying bug as well.

Impacted events are: `blur`, `clipboardchange`, `error`, `focus`, `gamepadconnected`, `gamepaddisconnected`, `load`, `securitypolicyviolation` (the last event should actually bubble but the extraction logic cannot detect that without some additional help).

### DIFF
--- a/src/postprocessing/events.js
+++ b/src/postprocessing/events.js
@@ -102,15 +102,12 @@ function expandMixinTargets(event, mixins) {
 
 
 function setBubblingPerTarget(event, parsedInterfaces) {
-  // if an event targets an interface in a tree
-  // but the root of the tree wasn't detected as a target
-  // we can assume bubbles is false
-  // (ideally, we should check the existence of the event handler on the
-  // root interface, but there is no easy way to get a consolidated IDL view
-  // of the root at the moment)
+  // If an event targets a non root interface in a tree, we can assume that
+  // bubbles is false (ideally, we should check the existence of the event
+  // handler on the root interface, but there is no easy way to get a
+  // consolidated IDL view of the root at the moment)
   if (!event.targets) return;
   const updatedTargets = [];
-  const detected = {};
   const treeInterfaces = [];
   for (let iface of event.targets) {
     const treeInfo = getInterfaceTreeInfo(iface, parsedInterfaces);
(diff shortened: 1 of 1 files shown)


tidoust:
I note we don't have tests for events consolidation, which obviously isn't fantastic...

dontcallmedom (approved):

PR #2091, opened 2026-05-21
Fix dfns extraction for RFC8610 (CDDL types)

tidoust:
Reffy creates a dfn for CDDL types defined in RFC8610.

A new rendition of RFC specs seems to have been put in place for `www.rfc-editor.org` URLs. On top of redirecting URLs, the new rendition adds a few unexpected HTML comments around section titles for some reason, and the anchose we used to detect the "Standard Prelude" appendix needed to be adjusted accordingly.

Detection is now done on actual text to avoid further problems. That remains imperfect, but will hopefully prove more stable over time.

### DIFF
--- a/src/browserlib/extract-dfns.mjs
+++ b/src/browserlib/extract-dfns.mjs
@@ -977,18 +977,21 @@ function preProcessSVG2() {
  */
 function preProcessRFC8610() {
   // The RFC is defined as a set of pages (yuck!)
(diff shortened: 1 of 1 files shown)


tidoust:
FYI, @dontcallmedom. Taking the liberty to merge without review, to restore CDDL type dfns in Webref, so that generation of specs with CDDL can resume! We may adjust afterwards.

PR #2101, opened 2026-06-05
Make algorithm extraction more resilient to invalid switch markup

tidoust:
Specs from the FIDO Alliance have algorithm-like steps that use of the `<dl class="switch">` form, but one of these only has `<dt>` without any `<dd>` which resulted in an error being thrown by Reffy.

The extraction logic can now handle that case, reporting the option without further details.


Elchi3 (approved):
Thanks @tidoust! Confirming this now crawls fido ctap 2.2.

(`npx reffy --spec fido-client-to-authenticator-protocol-v2.2`)

PR #2127, opened 2026-07-27
Add a `backrefs` post-processing module

tidoust:
This adds a new post-processing module to create `backrefs` extracts at the end of a crawl, as discussed in https://github.com/w3c/webref/issues/1934. The code reuses code from the Webref PR https://github.com/w3c/webref/pull/1997 by @DevOlabode, which in turn reuses code from Webdex.

The post-processing module runs at the crawl level. It creates one backrefs extract per spec that contains terms referenced by other specs. It is slightly unusual in that modules at the crawl level are usually meant to create something like a consolidated file, or completely different extracts (e.g., IDL names), but that seems to work just fine.

Easiest way to run the post-processing module given a path to Webref results and an output folder:

```bash
node reffy.js --use-crawl [path] --output [folder] --post backrefs
```

What's still missing:
- The JSON schema for the extract, which should probably go under `schemas/postprocessing`, even thought that's going to require 2 lines of logic in the `getSchemaValidationFunction` function in `util.js`.
- Ideally, a few tests to validate the module (but post-processing tests are a bit painful to write, and we may be able to survive without them to start with).

### DIFF
--- a/src/lib/post-processor.js
+++ b/src/lib/post-processor.js
@@ -51,6 +51,7 @@ import fs from 'node:fs';
 import path from 'node:path';
 import { pathToFileURL } from 'node:url';
 import { createFolderIfNeeded, shouldSaveToFile } from './util.js';
+import backrefs from '../postprocessing/backrefs.js';
 import csscomplete from '../postprocessing/csscomplete.js';
 import cssmerge from '../postprocessing/cssmerge.js';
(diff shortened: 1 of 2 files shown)


DevOlabode:
Thanks for putting this draft together. I took over the implementation as suggested and opened #2128

tidoust:
Closing in favor of #2128.

PR #2128, opened 2026-07-27
Add a backrefs post-processing module

DevOlabode:
## Summary
- Add a crawl-level `backrefs` post-processing module that builds per-spec back-reference extracts from curated `dfns` and `links` (terms defined in a spec that other specs reference).
- Reuses the algorithm from [w3c/webref#1997](https://github.com/w3c/webref/pull/1997) / Webdex, preserving document order (no `.sort()`).
- Adds the JSON schema under `schemas/postprocessing/backrefs.json`, wires it into `getSchemaValidationFunction`, and covers the module with unit tests.

This completes the work sketched in [w3c/reffy#2127](https://github.com/w3c/reffy/pull/2127) (schema + tests). Once released, Webref can drop `prepare-backrefs.js` and invoke the module via `crawlSpecs({ post: [..., 'backrefs'] })`.

Related: [w3c/webref#1934](https://github.com/w3c/webref/issues/1934)

Run against an existing crawl:

```bash
node reffy.js --use-crawl [path] --output [folder] --post backrefs
```

## Test plan
- [ ] `node --test test/generate-backrefs.js`
- [ ] Spot-check a generated extract (e.g. Streams) for external referrers only, and that term/referrer order matches dfn/document order
- [ ] Confirm schema validation succeeds for a sample `backrefs/*.json` via `getSchemaValidationFunction('backrefs')`

### DIFF
--- a/schemas/postprocessing/backrefs.json
+++ b/schemas/postprocessing/backrefs.json
@@ -21,10 +21,7 @@
           "linkingText": { "$ref": "../common.json#/$defs/strings" },
           "type": { "$ref": "../common.json#/$defs/dfnType" },
           "for": { "$ref": "../common.json#/$defs/strings" },
-          "access": {
-            "type": "string",
-            "enum": ["private", "public"]
-          },
+          "access": { "$ref": "../common.json#/$defs/access" },
           "referencedBy": {
             "type": "array",
             "minItems": 1,
--- a/schemas/postprocessing/backrefs.json
+++ b/schemas/postprocessing/backrefs.json
@@ -19,23 +19,7 @@
           "id": { "$ref": "../common.json#/$defs/id" },
           "href": { "$ref": "../common.json#/$defs/url" },
           "linkingText": { "$ref": "../common.json#/$defs/strings" },
-          "type": {
-            "type": "string",
-            "enum": [
-              "property", "descriptor", "value", "type",
-              "at-rule", "function", "selector",
-              "namespace", "interface", "constructor", "method", "argument",
-              "attribute", "callback", "dictionary", "dict-member", "enum",
-              "enum-value", "exception", "const", "typedef", "stringifier",
-              "serializer", "iterator", "maplike", "setlike", "extended-attribute",
-              "event", "permission",
-              "element", "element-state", "element-attr", "attr-value",
-              "cddl-module", "cddl-type", "cddl-parameter", "cddl-key", "cddl-value",
-              "scheme", "http-header",
-              "grammar", "abstract-op", "dfn"
-            ],
-            "$comment": "Types taken from src/browserlib/extract-dfns.mjs"
-          },
+          "type": { "$ref": "../common.json#/$defs/dfnType" },
           "for": { "$ref": "../common.json#/$defs/strings" },
           "access": {
             "type": "string",
--- a/schemas/postprocessing/backrefs.json
+++ b/schemas/postprocessing/backrefs.json
@@ -18,10 +18,7 @@
         "properties": {
           "id": { "$ref": "../common.json#/$defs/id" },
           "href": { "$ref": "../common.json#/$defs/url" },
-          "linkingText": {
-            "type": "array",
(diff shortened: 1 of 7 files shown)


schemas/postprocessing/backrefs.json
```diff
@@ -0,0 +1,68 @@
...
+      "description": "List of definitions that the spec owns and that are referenced by other specs, along with the specs that reference them.",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "additionalProperties": false,
+        "required": ["id", "href", "linkingText", "type", "for", "access", "referencedBy"],
+        "properties": {
+          "id": { "$ref": "../common.json#/$defs/id" },
+          "href": { "$ref": "../common.json#/$defs/url" },
+          "linkingText": {
```
tidoust: I should have done it in `browserlib/extract-dfns.json` already, but to DRY the schema a bit, I think we should create a `strings` type in the `common.json` file and reference it in both `browserlib/extract-dfns.json` and this schema through `{ "$ref": "../common.json#/$defs/strings" }` (for `linkingText`, `type` and `for`.

schemas/postprocessing/backrefs.json
```diff
@@ -0,0 +1,68 @@
...
+            "type": "array",
+            "items": { "type": "string" }
+          },
+          "type": {
```
tidoust: Same thing here, especially since the enumeration may evolve over time and needs to be the same for `dfns` extracts and `backrefs` extracts.

schemas/postprocessing/backrefs.json
```diff
@@ -0,0 +1,68 @@
...
+              "scheme", "http-header",
+              "grammar", "abstract-op", "dfn"
+            ],
+            "$comment": "Types taken from src/browserlib/extract-dfns.mjs"
+          },
+          "for": {
+            "type": "array",
+            "items": { "type": "string" }
+          },
+          "access": {
```
tidoust: Same thing, a shared `access` type definition in `common.json` would be better.

tidoust (changes requested):
Thanks! Just a few minor comments on the schema to make it easier to keep it in sync with the `dfns` one, but that looks good to me otherwise.

Running the code locally, the extracts take up 10.4MB, which remains reasonable.

@Elchi3, I'm wondering whether you have comments on the shape of the data in the `backrefs` extracts (see the JSON schema), notably:

- The definition keys that get included. The `href` key is all that's needed in theory if you have the `dfns` extracts at hand. This approach creates a more stand-alone set of extracts (also copying the `id`, `linkingText`, `for`, etc.). By construction, the data will be the same. Now, if we end up adding this data to `@webref/xref` or to another npm package, dfn information will be duplicated in the package, which may not be ideal (FWIW, that's ~1.3MB out of the 10.4MB). I don't feel strongly either way.
- Whether to include/exclude references within the defining spec itself. The current approach only lists references from *other* specs. Spec authoring tools typically already maintain the list of references from within the specs themselves (and they do it "live"), so don't need to know that a term is referenced by the very spec that defines it and will be only interested in this list of *other* specs. Same thing for a dashboard such as Webdex. If we included references within the defining spec, these tools would need to filter the list. Now, for analysis, it may be useful to know whether a term is referenced anywhere at all, including in the spec that defines it. That can be computed, but is slightly more involved as it requires looking into `links` extracts again. I'm thinking the "exclusion" approach is fine for now and avoids generating additional MB of data.

DevOlabode:
Thanks for the review! Addressed the schema DRY comments in three commits:

- Shared strings in common.json (used by linkingText / for, and localLinkingText in the dfns schema)
- Shared dfnType for the definition type enumeration
- Shared access for public / private

Both browserlib/extract-dfns.json and postprocessing/backrefs.json now reference those common defs.

tidoust (approved):
Many thanks, @DevOlabode! 

This looks all good to me, I'll merge the PR, release a new version of Reffy, and have Webref use it. Once done, we can look into actually creating the extracts in Webref.

PR #2131, opened 2026-07-29
Support alternate heading IDs from empty spans and oldids

### LINKED ISSUE #2088: Consider supporting alternate heading IDs listed in empty spans
WebAssembly specs have empty spans like `<span id="[id]"></span>` at the beginning of a section to create alternate IDs (probably to preserve historical ones?)

For example, the [Globals](https://webassembly.github.io/spec/core/bikeshed/#globals%E2%91%A0) section starts with:

```html
<section id="globals">
  <span id="syntax-global"></span>
  <span id="index-4③"></span>
  <h4 class="heading settled" data-level="2.5.4" id="globals①">
    <span class="secno">2.5.4. </span><span class="content">Globals</span><a class="self-link" href="#globals①"></a>
  </h4>
```

BCD may use these IDs to reference the spec. A general question here is whether to add these IDs to the data, or whether it would be preferable to update links that target them to use the main section ID. In the above example, `syntax-global` seems a "good" ID, while `index-4③` doesn't quite look like something that's going to remain stable, or more importantly that will always target the same section in the document.

Via https://github.com/mdn/browser-compat-data/pull/23958

DevOlabode:
## Summary
- Collect empty `<span id>` aliases placed before a section heading into `alternateIds` (WebAssembly/Bikeshed pattern such as `syntax-global`).
- For ECMAScript/Ecmarkup clauses, also include historical IDs from the `oldids` attribute and matching empty spans.
- Keep the section/clause ID as the primary heading `id`; empty-span / `oldids` values are alternates only.

This helps tools like BCD resolve older fragment links to the correct section.

Fixes #2088

## Test plan
- [ ] `PUPPETEER_CACHE_DIR="$HOME/.cache/puppeteer" node --test test/extract-headings.js`
- [ ] `PUPPETEER_CACHE_DIR="$HOME/.cache/puppeteer" node --test --test-name-pattern 'abstract operations from ecmascript|alternate heading IDs' test/extract-dfns.js`
- [ ] Spot-check a WebAssembly core headings extract for `globals` including `syntax-global` (and any `index-*` aliases) in `alternateIds`
- [ ] Spot-check an ECMAScript headings/dfns extract for an `emu-clause` with `oldids`

### DIFF
--- a/src/browserlib/extract-headings.mjs
+++ b/src/browserlib/extract-headings.mjs
@@ -11,13 +11,19 @@ export default function (spec, idToHeading) {
   const esHeadings = [...document.querySelectorAll('emu-clause[id] > h1')].map(n => {
     const headingNumber = n.querySelector(".secnum")?.textContent;
     const headingLevel = headingNumber ? headingNumber.split(".").length : undefined;
-    return {
-      id: n.parentNode.id,
-      href: getAbsoluteUrl(n.parentNode, { singlePage }),
+    const clause = n.parentNode;
+    const res = {
+      id: clause.id,
+      href: getAbsoluteUrl(clause, { singlePage }),
       title: n.textContent.replace(headingNumber, '').trim(),
       level: headingLevel,
       number: headingNumber
     };
+    const alternateIds = idToHeading[res.href]?.alternateIds;
+    if (alternateIds?.length) {
+      res.alternateIds = alternateIds;
+    }
+    return res;
   });
 
   // Headings using spans in www.rfc-editor.org RFCs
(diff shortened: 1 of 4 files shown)


tidoust:
Ah, I should do better at cleaning up the issues list and refreshing their status...

I haven't looked at the code in this PR but I suspect it would do the right thing. Now, the underlying issue was about pondering whether actually doing that was a good idea. The motivating use case was for BCD, but BCD now validates fragments using the data in `@webref/xref`, and had no real problem refreshing the "old" IDs to make that possible. Using more recent IDs also seems like a good thing to have in BCD.

Said differently, I think we should be driven by usage scenarios there before we update Reffy. The main usage scenario I had in mind (BCD) no longer needs it for now. If you or anyone else is planning to use this data, please share your use case!

DevOlabode:
Thanks for the context.

My motivation was primarily the original issue and making the extracted data more complete. 
Given that BCD no longer requires it and I don't currently have another concrete consumer, I'm happy to close the PR for now. 

If another use case emerges in the future, we can revisit it.

PR #2153, opened 2026-09-07
Detect nested headings linked to an element definition

tidoust:
Via https://github.com/w3c/browser-specs/issues/2255

The "Install Element" spec nests the heading that defines the `<install>` element, while the code expected the heading to always be at the same level as the definition list that describes the element.

This update makes the code more flexible when it looks for the heading.

Note: I ran Reffy locally across all specs to confirm that it wouldn't create any false positive.


Elchi3 (approved):
Thanks @tidoust! 👍

PR #2163, opened 2026-09-16
Save file extracts when current spec in the series is not crawled

tidoust:
For CSS and IDL extracts, Reffy produces only one file extract per spec series, which is typically the file extract for the *current specification* in the series, or more accurately for the latest (and fullest) level in the series that defines CSS or IDL, and is either the current specification or a more recent level.

This logic works well when Reffy is told to crawl all specs in browser-specs because all series have a current specification in the list. But Reffy may also be run on a single spec or a subset of the full list, and there's no guarantee that this subset will contain the current specification in the series.

This update makes Reffy skip the "more recent" check in the case when the current specification is not in the list of crawled specs. This typically makes it possible to run Reffy on an individual spec such as `css-cascade-3` and get the actual extract files from it, even though the CSS extract would be skipped if the current spec (`css-cascade-4`) was also crawled.


Elchi3 (approved):

src/lib/util.js
```diff
@@ -759,7 +759,11 @@ function isLatestLevelThatPasses(spec, list, predicate) {
     }
 
     // Make sure that spec is the current one or is more recent than the
-    // current one.
+    // current one. If the list of crawled specs does not contain the current
+    // specification, we'll just assume that the spec is recent enough.
+    if (!list.find(s => s.shortname === spec.series.currentSpecification)) {
+        return true;
+    }
```
deniak, line 766: IIUC, when the list doesn't contain the current spec but still the specs with the other level, e.g. `css-cascade-3`, `css-cascade-5` and `css-cascade-6` (`css-cascade-4` being current), then `isLatestLevelThatPasses` will return all 3 specs. Shouldn't it return the newest level only?

tidoust: It will not return `css-cascade-6` because it's a `delta` spec. But you're right, it would indeed return `css-cascade-3` **and** `css-cascade-5`. And one would override the other in the end (I don't even know which one, it probably depends on the order in which the specs get listed in the first place).

Now, without the current spec in the series, there's no good way to select the right extract (it's obvious for us but nothing mandates that `css-cascade-5` comes **after** `css-cascade-3`, the code follows previous/next links to determine the relative order in a series, and without `css-cascade-4`, the code just knows that the specs are in the same series).

I'm not sure what people would expect if they crawled `css-cascade-3` and `css-cascade-5` without `css-cascade-4`, probably to get all extract files, and not only the one from `css-cascade-5`. My hunch is that we shouldn't have made Reffy create extracts per series in the first place and leave that consolidation to the curation phase in Webref, but @dontcallmedom is even more stubborn than me ;)

I'll record that as an issue but I think it's fine to leave it as "broken" for now. The use case that I was really trying to fix here is when you run Reffy on a single spec such as `css-cascade-3` or `SVG11`, and don't get the corresponding CSS/IDL extracts.

PR #2166, opened 2026-09-18
Exclude CDDL blocks in the whole Bikeshed CDDL index section

titusfortner:
## Problem

`extract-cddl.mjs` intends to skip the "CDDL Index" appendix that Bikeshed generates, but the exclusion is implemented as `closest('#cddl-index')`. Bikeshed emits `<h2 id="cddl-index">` followed by sibling `<h3>` and `<pre class="cddl">` elements, so the generated blocks are never descendants of the element carrying the ID and the exclusion never matches anything.

As a result, any Bikeshed spec with a populated CDDL index gets its definitions extracted twice: once from the source blocks and once from the index. When the spec defines CDDL modules, the index blocks carry no `data-cddl-module`, so they also leak into every per-module extract. Specs that only reference modules defined elsewhere get an empty index from Bikeshed and are unaffected.

## Solution

- `extract-cddl.mjs` excludes every `pre.cddl` in the `#cddl-index` section, which runs from that heading to the next heading of the same or higher level in document order. This handles `<section>` wrapping and the generated per-module `<h3>` subsections.
- `getCodeElements()` gains an additive `excludeFilter` predicate, called on the live element before cloning, since section membership cannot be tested on the cleaned clones it returns. `excludeSelectors` keeps its ancestor semantics.
- Verified on the current editor's drafts of every Bikeshed spec with CDDL extracts in webref: WebDriver BiDi and AT Driver keep identical rule-name sets with each rule now appearing once; prefetch, UA client hints, and Web Bluetooth are byte-identical before and after.


tidoust (approved):
Oh, indeed. I had missed that last year when Bikeshed started to support CDDL blocks. Bikeshed prefers to keep things flat where ReSpec relies on sectioning...

Thanks for the pull request! Looks good to me.

PR #2167, opened 2026-09-18
Strip the header that ReSpec adds to CDDL blocks

titusfortner:
## Problem

ReSpec now processes every `pre.cddl` block and prepends a `<span class="cddlHeader">` holding a "CDDL" self-link, the same way it adds an `idlHeader` to WebIDL blocks. `clone-and-clean.mjs` strips `.idlHeader` but not `.cddlHeader`, so the first rule of every block comes out prefixed with `CDDL`.

This affects any ReSpec spec with CDDL once it is built with a current ReSpec. In webref today that is Digital Credentials, whose extracts have `CDDLdigitalCredentials.VirtualWalletAction` and two more rules like it. Permissions is unaffected only because its published snapshot predates the ReSpec CDDL module.

## Solution

- Adds `.cddlHeader` to `informative-selector.mjs` next to `.idlHeader`. The header is the only thing ReSpec's CDDL module injects that is not part of the CDDL text, and with this change the extractor reproduces the Digital Credentials source CDDL exactly.
- Adds a test with the markup ReSpec generates, matching the existing `idlHeader` case in `extract-webidl.js`.


tidoust (approved):
Thanks!

I was wondering why CDDL validation in Webref didn't flag the affected CDDL extracts, but that `CDDL` just ended up being the prefix of the root rule, which doesn't create an invalid name per se.

